PR #112, opened 2023-08-10
Unity Support

ndesai-newrelic:
1) Unity Custom Exception
2) Unity Support


ywang-nr (approved):

agent-core/src/main/java/com/newrelic/agent/android/Agent.java
```diff
@@ -17,6 +17,7 @@
 public class Agent {
     public static final String VERSION = "#VERSION#";
     public static final String MONO_INSTRUMENTATION_FLAG = "#MONO_INSTRUMENTATION_FLAG#";
+    public static final String UNITY_INSTRUMENTATION_FLAG = "#UNITY_INSTRUMENTATION_FLAG#";
```
cthomas-newrelic, line 20: How is "_Unity Instrumented_" different than Mono? Is Mono still used? If not remove, or reprovision that? What other agent would use either flag?

agent-core/src/main/java/com/newrelic/agent/android/Agent.java
```diff
@@ -211,4 +212,8 @@ public static boolean isInstantApp() {
         return getImpl().isInstantApp();
     }
 
+    public static String getUnityInstrumentationFlag() {
```
cthomas-newrelic, line 215: Who/what calls this?

agent-core/src/main/java/com/newrelic/agent/android/ApplicationFramework.java
```diff
@@ -12,5 +12,7 @@ public enum ApplicationFramework {
     Flutter,
     Capacitor,
     Xamarin,
-    MAUI
+    MAUI,
+
```
cthomas-newrelic, line 16: Remove blank line, and make it caps-case.

cthomas-newrelic (changes requested):
I feel like since Unity was the only hybrid platform we supported in the past, this seems redundant. Or was it all removed during OSS migration?

Please move to agent-core/src/main/java/com/newrelic/agent/android/hybrid/unity

agent-core/src/main/java/com/newrelic/agent/android/Agent.java
ndesai-newrelic: removed this

agent-core/src/main/java/com/newrelic/agent/android/ApplicationFramework.java
cthomas-newrelic: Should `Maui`. It's not an acronym is it?

agent-core/src/main/java/com/newrelic/agent/android/Agent.java
cthomas-newrelic: I still don't see an answer for this. The argument for using both is we have another hybrid target that is also Mono-based. MONO_INSTRUMENTATION_FLAG was added explicitly for Unity (not Xamarin), which is why it's redundant.

Also, it doesn't appear the build files were modified for `UNITY_INSTRUMENTATION_FLAG`, so simply adding another flag isn't enough.

gradle.properties
```diff
@@ -9,7 +9,7 @@ android.useAndroidX=true
 android.disableAutomaticComponentCreation=true
 
 # Version of the SDK to be built and deployed
-newrelic.agent.version = 7.0.0
+newrelic.agent.version = 7.0.1
```
cthomas-newrelic, line 12: Don't bump the version in a branch, unless it's a `release` branch or explicitly targeted for a release. There won't be a 7.0.1 release, btw.

cthomas-newrelic (changes requested):
Doesn't seem much different than the previous review.  Any code you see in the agent relating to Mono was for Unity - why its is redundant. With probably a few exceptions, when you restored MONO support to the agent you were in effect restoring Unity support. 

If MONO is too general to be used for hybrid implementations, then remove it rather than Unity references.

gradle.properties
ndesai-newrelic: i did these changes to test sonatype for Unity changes.

agent-core/src/main/java/com/newrelic/agent/android/Agent.java
ndesai-newrelic: @cthomas-newrelic you are right, those changes are not required.

gradle.properties
cthomas-newrelic: Yea, I figured as much. I also accidentally commit the properties file

cthomas-newrelic (approved):
So now, is the MONO flag still needed, even for Xamarin/Maui?

PR #120, opened 2023-09-12
[NR-154228] fix: excludeVariant extension API wasn't honoring case of provided variant

cthomas-newrelic:
The fix was a one-liner, and most work is an effort to get plugin integration tests back online. Integration tests are much faster then the functional tests, so the dev turnaround time is lower. Still trying to address the Gradle Test runner OOM conditions, however. 

Additional logging was added to plugin startup to provide an audit of what variants are being excluded.


plugins/gradle/src/main/groovy/com/newrelic/agent/android/NewRelicExtension.groovy
```diff
@@ -146,7 +150,7 @@ abstract class NewRelicExtension {
         variantConfigurations.findByName(variantName.toLowerCase())?.with {
             return instrument
         }
-        return !variantExclusions.findAll { variantName == it || variantName.endsWith(it.capitalize()) }.empty
+        return !variantExclusions.findAll { variantName.toLowerCase() == it || variantName.endsWith(it.capitalize()) }.empty
```
cthomas-newrelic, line 153: This was the cause of the failure

ndesai-newrelic (approved):

PR #132, opened 2023-09-27
[NR-163742] AGP 7.2, 7.3 must use legacy class transformer

cthomas-newrelic:
Transform API is still available to those using Gradle 7.4.0 and below, so default to that for AGP 7.3, 7.2 and lower.

### DIFF
--- a/plugins/gradle/src/main/groovy/com/newrelic/agent/android/VariantAdapter.groovy
+++ b/plugins/gradle/src/main/groovy/com/newrelic/agent/android/VariantAdapter.groovy
@@ -62,17 +62,13 @@ abstract class VariantAdapter {
      */
     static VariantAdapter register(BuildHelper buildHelper) {
         def currentGradleVersion = GradleVersion.version(buildHelper.getGradleVersion())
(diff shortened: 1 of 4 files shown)


plugins/gradle/src/main/groovy/com/newrelic/agent/android/VariantAdapter.groovy
```diff
@@ -62,17 +62,13 @@ abstract class VariantAdapter {
...
 
-        if (currentGradleVersion >= GradleVersion.version("7.4")) {
-            return new AGP74Adapter(buildHelper)
-        } else if (currentGradleVersion >= GradleVersion.version("7.2")) {
-            if (GradleVersion.version(buildHelper.agpVersion) < GradleVersion.version("7.2")) {
-                return new AGP4Adapter(buildHelper)
-            }
-            return new AGP70Adapter(buildHelper)
-        } else {
+        if (currentAgpVersion < GradleVersion.version("7.4")) {
```
cthomas-newrelic, line 67: Here we'll only check the AGP version and assume customer has proper corresponding Gradle level.

ywang-nr (approved):

PR #135, opened 2023-10-03
chore: add script to bulk delete GHA workflow runs

cthomas-newrelic:
Also adds adds workflow job to be run at the beginning of each month.

NOTE: script is set for _dryrun_, meaning no data will be deleted until all changes are verified.


ywang-nr (approved):
Great idea!

ywang-nr (approved):

PR #140, opened 2023-10-10
[NR-168456] Exception handling around classTransformer

ywang-nr:
Bug fix: https://issues.newrelic.com/browse/NR-168456

### DIFF
--- a/plugins/gradle/src/main/groovy/com/newrelic/agent/android/ClassTransformWrapperTask.groovy
+++ b/plugins/gradle/src/main/groovy/com/newrelic/agent/android/ClassTransformWrapperTask.groovy
@@ -7,7 +7,6 @@ package com.newrelic.agent.android
 
 import com.newrelic.agent.compile.ClassTransformer
 import groovy.io.FileType
-import org.apache.commons.io.IOUtils
 import org.gradle.api.DefaultTask
 import org.gradle.api.file.Directory
 import org.gradle.api.file.DirectoryProperty
@@ -65,11 +64,11 @@ abstract class ClassTransformWrapperTask extends DefaultTask {
                                             jarOutputStream << it
                                         }
                                     } catch (RuntimeException re) {
-                                        logger.warn("[ClassTransformTask] encountered a RuntimeException: " + re.getMessage() + ", output the original jar file")
-                                        IOUtils.copy(fileInputStream, jarOutputStream)
+                                        logger.warn("[ClassTransformTask] Instrumentation is disabled for " + classFile.name + " with exception: "+ re.getLocalizedMessage())
+                                        jarOutputStream << fileInputStream
                                     } catch (IOException ioE) {
-                                        logger.warn("[ClassTransformTask] encountered a IOException: " + ioE.getMessage() + ", output the original jar file")
-                                        IOUtils.copy(fileInputStream, jarOutputStream)
+                                        logger.warn("[ClassTransformTask] Instrumentation is disabled for " + classFile.name + " with exception: "+ ioE.getLocalizedMessage())
+                                        jarOutputStream << fileInputStream
                                     }
                                     jarOutputStream.closeEntry()
                                 } catch (IOException ioE) {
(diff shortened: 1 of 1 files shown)


plugins/gradle/src/main/groovy/com/newrelic/agent/android/ClassTransformWrapperTask.groovy
```diff
@@ -59,8 +60,16 @@ abstract class ClassTransformWrapperTask extends DefaultTask {
...
                                 try {
                                     jarOutputStream.putNextEntry(jarEntry)
-                                    transformer.processClassBytes(classFile, fileInputStream).withCloseable {
-                                        jarOutputStream << it
+                                    try {
+                                        transformer.processClassBytes(classFile, fileInputStream).withCloseable {
+                                            jarOutputStream << it
+                                        }
+                                    } catch (RuntimeException re) {
+                                        logger.warn("[ClassTransformTask] encountered a RuntimeException: " + re.getMessage() + ", output the original jar file")
```
cthomas-newrelic, line 68: How about "_instrumentation is disabled for <class file>_"? Also, don;t seend to say both "_encountered a RuntimeException_" and "_re.getMessage()_". The exception message is sufficient.

plugins/gradle/src/main/groovy/com/newrelic/agent/android/ClassTransformWrapperTask.groovy
```diff
@@ -59,8 +60,16 @@ abstract class ClassTransformWrapperTask extends DefaultTask {
...
+                                        IOUtils.copy(fileInputStream, jarOutputStream)
```
cthomas-newrelic, line 69: Just stream the input to the output: ```jarOutputStream << fileInputStream;```

Also, if there's a problem during the stream read, we may only get part of it back. ot much we can do other than buffer the class bytes.

cthomas-newrelic (changes requested):

plugins/gradle/src/main/groovy/com/newrelic/agent/android/ClassTransformWrapperTask.groovy
```diff
@@ -7,6 +7,7 @@ package com.newrelic.agent.android
 
 import com.newrelic.agent.compile.ClassTransformer
 import groovy.io.FileType
+import org.apache.commons.io.IOUtils
```
cthomas-newrelic, line 10: Try to avoid using Apache commons. I've been trying to wean off of it entirely.

plugins/gradle/src/main/groovy/com/newrelic/agent/android/ClassTransformWrapperTask.groovy
ywang-nr: Does this sound better/
`logger.warn("[ClassTransformTask] Instrumentation is disabled for " + classFile.name + " with exception: "+ re.getMessage())`

cthomas-newrelic (approved):
Now we have code coverage checks, and tis could be a touch case to test. Can the c/c checks be disabled for now? Getting a reproduction case will inform adding useful tests.

Squash please.

plugins/gradle/src/main/groovy/com/newrelic/agent/android/ClassTransformWrapperTask.groovy
cthomas-newrelic: Remove unused imports

ywang-nr: Yes, it's removed. This is an outdated file.

ywang-nr:
> Now we have code coverage checks, and tis could be a touch case to test. Can the c/c checks be disabled for now? Getting a reproduction case will inform adding useful tests.
> 
> Squash please.

I could disable codecov for plugin if we don't want it.

PR #142, opened 2023-10-24
Mono build github action workflow and build script

ndesai-newrelic:


.github/workflows/sonatypeStaticDeployment.yml
```diff
@@ -0,0 +1,48 @@
...
+    steps:
+    - uses: actions/checkout@v3
+    - name: Setup JDK 11
+      uses: actions/setup-java@v3
+      with:
+          java-version: '11'
+          distribution: 'temurin'
+          cache: 'gradle'
+    - name: Build with Gradle
+      run: ./gradlew clean assemble --no-daemon -PmonoEnabled=true
```
ywang-nr, line 26: Is it a good idea using if statement to combine these two sonatype deployment since monoEnabled=true is the only difference?

ndesai-newrelic: we can do that, let me do change for that

ndesai-newrelic:
@ywang-nr i added publish mono artifact in the the same deplotyment file

cthomas-newrelic (approved):
🐑 it

cthomas-newrelic:
Please SQUASH MERGE and provide a good summary and description of the change. I just noticed it was 4 commits with teh same messages

PR #143, opened 2023-10-24
HttpHeaders as Custom Attributes and GRAPHQL Request Headers Capture

ndesai-newrelic:
1) implemented a new feature to track HTTP headers that are added by the client when making an HTTP request. To achieve this, we added a static method that records these headers and adds them as custom attributes at the HTTP request level.

2) added a new feature to capture Apollo GraphQL HTTP headers, specifically related to the operation name, type, and ID. By capturing these headers, we can provide more granular insights into how GraphQL is being used, empowering our customers to gain a deeper understanding of their GraphQL data.

### DIFF
--- a/agent-core/src/main/java/com/newrelic/agent/android/HttpHeaders.java
+++ b/agent-core/src/main/java/com/newrelic/agent/android/HttpHeaders.java
@@ -1,0 +1,62 @@
+package com.newrelic.agent.android;
+
+
+import com.newrelic.agent.android.util.Constants;
+
+import java.util.HashSet;
+import java.util.List;
+import java.util.Set;
+import java.util.concurrent.atomic.AtomicReference;
+
+public class HttpHeaders {
+
+    protected static final AtomicReference<HttpHeaders> instance = new AtomicReference<>(new HttpHeaders());
+    private final Set<String> httpHeaders;
+
+    public static final String OPERATION_NAME = "operationName";
+    public static final String OPERATION_TYPE = "operationType";
+    public static final String OPERATION_ID = "operationId";
+
+    private HttpHeaders() {
+        httpHeaders = new HashSet<>();
+        httpHeaders.add(Constants.ApolloGraphQLHeader.OPERATION_NAME);
+        httpHeaders.add(Constants.ApolloGraphQLHeader.OPERATION_ID);
+        httpHeaders.add(Constants.ApolloGraphQLHeader.OPERATION_TYPE);
+    }
+
+    public static HttpHeaders getInstance() {
+        return instance.get();
+    }
+
+    public void addHttpHeaderAsAttribute(String httpHeader) {
+        httpHeaders.add(httpHeader);
+    }
+
+    public void removeHttpHeaderAsAttribute(String httpHeader) {
+        if (!(Constants.ApolloGraphQLHeader.OPERATION_ID.equalsIgnoreCase(httpHeader) || Constants.ApolloGraphQLHeader.OPERATION_NAME.equalsIgnoreCase(httpHeader) || Constants.ApolloGraphQLHeader.OPERATION_TYPE.equalsIgnoreCase(httpHeader)))
+            httpHeaders.remove(httpHeader);
+    }
+
+    public boolean addHttpHeadersAsAttributes(List<String> httpHeaders) {
+        return this.httpHeaders.addAll(httpHeaders);
+    }
+
+    public Set<String> getHttpHeaders() {
+        return httpHeaders;
+    }
+
+
+    public static String translateApolloHeader(String s) {
+
+        switch (s) {
+            case Constants.ApolloGraphQLHeader.OPERATION_NAME:
+                return OPERATION_NAME;
+            case Constants.ApolloGraphQLHeader.OPERATION_ID:
+                return OPERATION_ID;
+            case Constants.ApolloGraphQLHeader.OPERATION_TYPE:
+                return OPERATION_TYPE;
+            default:
+                return s;
+        }
+    }
+}
--- a/agent-core/src/main/java/com/newrelic/agent/android/instrumentation/HttpURLConnectionExtension.java
+++ b/agent-core/src/main/java/com/newrelic/agent/android/instrumentation/HttpURLConnectionExtension.java
@@ -5,7 +5,6 @@
 
 package com.newrelic.agent.android.instrumentation;
 
-import com.newrelic.agent.android.Measurements;
 import com.newrelic.agent.android.TaskQueue;
 import com.newrelic.agent.android.api.common.TransactionData;
 import com.newrelic.agent.android.instrumentation.io.CountingInputStream;
@@ -40,7 +39,7 @@ import java.util.TreeMap;
 // but I don't think it should otherwise affect us.
 //
 public class HttpURLConnectionExtension extends HttpURLConnection {
-    private HttpURLConnection impl;
+    private final HttpURLConnection impl;
     private TransactionState transactionState;
     private CountingInputStream errorStream = null;
 
@@ -59,6 +58,7 @@ public class HttpURLConnectionExtension extends HttpURLConnection {
     @Override
     public void addRequestProperty(String field, String newValue) {
         impl.addRequestProperty(field, newValue);
+        TransactionStateUtil.addHeadersAsCustomAttribute(transactionState, field, newValue);
     }
 
     @Override
@@ -186,7 +186,7 @@ public class HttpURLConnectionExtension extends HttpURLConnection {
                 errorStream = new CountingInputStream(impl.getErrorStream(), true);
             }
         } catch (Exception e) {
-            log.error("HttpsURLConnectionExtension: " + e.toString());
+            log.error("HttpsURLConnectionExtension: " + e);
             return impl.getErrorStream();
         }
         return errorStream;
@@ -496,6 +496,7 @@ public class HttpURLConnectionExtension extends HttpURLConnection {
     @Override
     public void setRequestProperty(String field, String newValue) {
         impl.setRequestProperty(field, newValue);
+        TransactionStateUtil.addHeadersAsCustomAttribute(transactionState, field, newValue);
     }
 
     @Override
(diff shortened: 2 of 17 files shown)


agent-core/src/main/java/com/newrelic/agent/android/HttpHeaders.java
```diff
@@ -0,0 +1,47 @@
...
+            instance = new HttpHeaders();
+        }
+        return instance;
+    }
+
+    public void addHttpHeaderAsAttribute(String httpHeader) {
+        httpHeaders.add(httpHeader);
+    }
+
+    public void removeHttpHeaderAsAttribute(String httpHeader) {
```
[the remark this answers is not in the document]
ndesai-newrelic: I added the removeHttpHeaderAsAttribute method to give our customers the flexibility to remove specific HTTP headers at runtime for particular HTTP requests. This allows them to prevent the headers from being captured if they don't want to include them for specific purposes.

agent-core/src/main/java/com/newrelic/agent/android/instrumentation/HttpURLConnectionExtension.java
```diff
@@ -576,14 +577,17 @@ void addTransactionAndErrorData(TransactionState transactionState) {
...
+            /* Also add Content-Length. TransactionState bytesReceived is set directly
              * from the Content-Length header in this error case.
-			*/
+             */
             params.put(Constants.Transactions.CONTENT_LENGTH, transactionState.getBytesReceived() + "");
 
             transactionData.setResponseBody(responseBody);
-            transactionData.setParams(params);
-
+            if (transactionData.getParams() != null) {
```
cthomas-newrelic, line 586: Why not just alloc the `params` field in the ctor, rather than duplicate this nullability check each place it's used?

agent-core/src/main/java/com/newrelic/agent/android/instrumentation/TransactionState.java
```diff
@@ -229,7 +240,7 @@ TransactionData toTransactionData() {
                     bytesSent,
                     bytesReceived,
                     appData,
-                    wanType, trace, null);
+                    wanType, trace,"",params, null);
```
cthomas-newrelic, line 243: Formatting. Don't forget to clean before commit. What does the parameter `""` represent? Is there a better way to communicate this (such as using a string constant)?

agent-core/src/main/java/com/newrelic/agent/android/instrumentation/okhttp2/OkHttp2Instrumentation.java
```diff
@@ -55,6 +60,17 @@ public static Call newCall(OkHttpClient client, Request request) {
         return new CallExtension(client, request, client.newCall(request), transactionState);
     }
 
+    private static void addHeadersAsCustomAttribute(TransactionState transactionState, Request request) {
+
+        Map<String, String> headers = new HashMap<>();
+        for (String s : HttpHeaders.getInstance().getHttpHeaders()) {
+            if (request.headers().get(s) != null) {
```
cthomas-newrelic, line 67: If `getHttpHeaders()` was guaranteed to return a non-null collection with non-null elements, this nullability check would not be needed.

agent-core/src/main/java/com/newrelic/agent/android/instrumentation/okhttp2/OkHttp2TransactionStateUtil.java
```diff
@@ -132,8 +132,11 @@ protected static Response addTransactionAndErrorData(TransactionState transactio
                 }
 
                 transactionData.setResponseBody(responseBodyString);
-                transactionData.setParams(params);
-
+                if (transactionData.getParams() != null) {
```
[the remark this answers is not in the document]
cthomas-newrelic: Again, alloc `params` in ctor then this conditional branch is no necessary (you can always add to the collection).

agent-core/src/main/java/com/newrelic/agent/android/instrumentation/okhttp3/OkHttp3Instrumentation.java
```diff
@@ -58,6 +63,17 @@ public static Call newCall(OkHttpClient client, Request request) {
         return new CallExtension(client, request, client.newCall(request), transactionState);
     }
 
+    private static void addHeadersAsCustomAttribute(TransactionState transactionState, Request request) {
+
+        Map<String, String> headers = new HashMap<>();
+        for (String s : HttpHeaders.getInstance().getHttpHeaders()) {
+            if (request.headers().get(s) != null) {
+                headers.put(Constants.ApolloGraphQLHeader.normalizeApolloHeader(s), request.headers().get(s));
```
cthomas-newrelic, line 71: This is really confusing. The method seems general enough, but uses a `ApolloGraphQLHeader()`* constant static method* to normalize it? 

You can also call the `request.headers.contains(s)` method as well and avoid null checks.

agent-core/src/main/java/com/newrelic/agent/android/instrumentation/okhttp3/OkHttp3TransactionStateUtil.java
```diff
@@ -158,7 +157,11 @@ protected static Response addTransactionAndErrorData(TransactionState transactio
                 }
 
                 transactionData.setResponseBody(responseBodyString);
-                transactionData.setParams(params);
+                if (transactionData.getParams() != null) {
```
cthomas-newrelic, line 160: See above.

agent-core/src/main/java/com/newrelic/agent/android/util/Constants.java
```diff
@@ -12,6 +12,30 @@ public final class Transactions {
...
+    public static final class ApolloGraphQLHeader {
+        public static final String OPERATION_NAME = "X-APOLLO-OPERATION-NAME";
+        public static final String OPERATION_TYPE = "X-APOLLO-OPERATION-TYPE";
+        public static final String OPERATION_ID = "X-APOLLO-OPERATION-ID";
+
+        public static final String OPERATION_NORMALIZE_NAME = "operationName";
+        public static final String OPERATION_NORMALIZE_TYPE = "operationType";
+        public static final String OPERATION_NORMALIZE_ID = "operationId";
+
+        public static String normalizeApolloHeader(String s) {
```
cthomas-newrelic, line 24: I get your intention, but no code with the constants, please.

agent-core/src/main/java/com/newrelic/agent/android/HttpHeaders.java
```diff
@@ -0,0 +1,47 @@
...
+import static com.newrelic.agent.android.util.Constants.ApolloGraphQLHeader.OPERATION_NAME;
+import static com.newrelic.agent.android.util.Constants.ApolloGraphQLHeader.OPERATION_TYPE;
+
+import java.util.HashSet;
+import java.util.List;
+import java.util.Set;
+
+public class HttpHeaders {
+
+    private static HttpHeaders instance = null;
```
cthomas-newrelic, line 13: If you're using a singleton, follow the pattern used elsewhere in the agent and use an [AtomIcReference<T>](https://docs.oracle.com/javase/8/docs/api/java/util/concurrent/atomic/AtomicReference.html).

agent-core/src/main/java/com/newrelic/agent/android/instrumentation/TransactionState.java
```diff
@@ -46,6 +47,16 @@ enum State {
     private TransactionData transactionData;
     private TraceContext trace;
 
+    private Map<String, String> params = null;
```
cthomas-newrelic, line 50: As mentioned, the container can be initialized here, or allocated in the ctor, and then avoid null checks when accessing it.

agent-core/src/main/java/com/newrelic/agent/android/util/Constants.java
cthomas-newrelic: This method seems general enough to avoid an `Apollo` reference. Should probably be a method in `HttpHeaders`.

cthomas-newrelic: It's petty, but the operation s really more a translation than a normalizaton.

cthomas-newrelic (changes requested):
I feel like this commit is really two distinct issues: header attributes and GraphQL instrumentation. I understand how closely they are related in this case, but maybe in the future break up your commits: the first commit would add the header support, then a follow up commit to add GraphQL headers. Make each commit more atomic. Use squash commits when you have several trivial commits thatcould have been done in one.

agent-core/src/main/java/com/newrelic/agent/android/instrumentation/TransactionState.java
ndesai-newrelic: i will do it.

agent-core/src/main/java/com/newrelic/agent/android/instrumentation/HttpURLConnectionExtension.java
ndesai-newrelic: got it.

agent-core/src/main/java/com/newrelic/agent/android/instrumentation/TransactionState.java
ndesai-newrelic: this responsebody value is set by setter method and Transcation state does not have response body, i am passing empty value . we can add constant value for empty string.

agent-core/src/main/java/com/newrelic/agent/android/instrumentation/okhttp2/OkHttp2Instrumentation.java
ndesai-newrelic: request.headers().get(s) can returns null and if we dont have any headers to add as custom attribute in that case it returns null.

agent-core/src/main/java/com/newrelic/agent/android/util/Constants.java
ndesai-newrelic: got this, will do necessary changes.

cthomas-newrelic (approved):
Please squash commit. 🚢 it

ywang-nr:
Quick question: what is ticket number for this PR? I can link it to our next release.

ndesai-newrelic:
@ywang-nr NR-147494

PR #148, opened 2023-11-03
[NR-172210] Event Persistence

ywang-nr:
[Ticket](https://new-relic.atlassian.net/browse/NR-172210): Events are stored in memory, in the event of a force kill of the app, all events will be lost.
1. Add SharedPreference to store events
2. Retrieve events when re-initialize the eventManager

### DIFF
--- a/agent-core/src/main/java/com/newrelic/agent/android/analytics/EventManagerImpl.java
+++ b/agent-core/src/main/java/com/newrelic/agent/android/analytics/EventManagerImpl.java
@@ -40,6 +40,8 @@ public class EventManagerImpl implements EventManager, EventListener {
     private final AtomicReference<EventListener> listener = new AtomicReference<EventListener>(this);
     AtomicReference<EventManager> instance = new AtomicReference<>(null);
 
+    private AnalyticsEventStore eventStore;
+
     public EventManagerImpl() {
         // Currently using 1000 max events, 5 minutes max buffer age
         this(DEFAULT_MAX_EVENT_BUFFER_SIZE, DEFAULT_MAX_EVENT_BUFFER_TIME);
@@ -59,8 +61,17 @@ public class EventManagerImpl implements EventManager, EventListener {
 
     @Override
     public void initialize(AgentConfiguration agentConfiguration) {
+        eventStore = agentConfiguration.getEventStore();
+        List<AnalyticsEvent> storedEvents = eventStore.fetchAll();
+
         if (!initialized.compareAndSet(false, true)) {
+            eventsRecorded.set(0);
+            eventsEvicted.set(0);
             log.verbose("EventManagerImpl.initialize(): Has already been initialized. Bypassing...");
+            //retrieve all the events from pref
+            for (AnalyticsEvent event : storedEvents) {
+                addEvent(event);
+            }
             return;
         }
 
@@ -68,6 +79,11 @@ public class EventManagerImpl implements EventManager, EventListener {
         eventsRecorded.set(0);
         eventsEvicted.set(0);
         empty();
+
+        //retrieve all the events from pref
+        for (AnalyticsEvent event : storedEvents) {
+            addEvent(event);
+        }
 
         listener.get().onStart(this);
     }
@@ -184,6 +200,7 @@ public class EventManagerImpl implements EventManager, EventListener {
 
             if (events.get().add(event)) {
                 // log.audit("Event added: [" + event.asJson() + "]");
+                eventStore.store(event);
                 eventsRecorded.incrementAndGet();
                 return true;
             }
@@ -363,7 +380,7 @@ public class EventManagerImpl implements EventManager, EventListener {
     /**
      * Notified when event buffer flush time has been exceeded. This normally occurs
      * between harvest cycles.
-     *
+     * <p>
      * The caller can update the limit using @link NewRelic.setMaxEventBufferTime()
      *
      * @param maxBufferTimeInSec
--- a/agent/src/main/java/com/newrelic/agent/android/stores/SharedPrefsEventStore.java
+++ b/agent/src/main/java/com/newrelic/agent/android/stores/SharedPrefsEventStore.java
@@ -1,0 +1,85 @@
+/*
+ * Copyright (c) 2022-present New Relic Corporation. All rights reserved.
+ * SPDX-License-Identifier: Apache-2.0
+ */
+
+package com.newrelic.agent.android.stores;
+
+import android.content.Context;
+import android.content.SharedPreferences;
+
+import com.google.gson.Gson;
+import com.google.gson.JsonObject;
+import com.newrelic.agent.android.analytics.AnalyticsEvent;
+import com.newrelic.agent.android.analytics.AnalyticsEventStore;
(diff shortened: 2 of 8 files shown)


agent/src/main/java/com/newrelic/agent/android/stores/SharedPrefsEventStore.java
```diff
@@ -0,0 +1,85 @@
...
+        synchronized (this) {
+            try {
+                final JsonObject jsonObj = event.asJsonObject();
+                String eventJson = jsonObj.toString();
+
+                // events should be stored synchronously, since the app is terminating
+                SharedPreferences.Editor editor = this.sharedPrefs.edit();
+                editor.putString(String.valueOf(event.getTimestamp()), eventJson);
+
+//                StatsEngine.get().inc(MetricNames.SUPPORTABILITY_CRASH_SIZE_UNCOMPRESSED, eventJson.length());
```
ywang-nr, line 49: Remove or update this comment

ywang-nr:
SharedPrefsEventStoreTest tests passed individually but not at build time.

cthomas-newrelic (changes requested):
Yes, using a munged key here is the what's needed, however I still don't tis provides adequate uniqueness. At this point, I think it's time to add a UUID to the AnalyticsEvent class itself (like we do for other payload types). That makes this issue moot, but be careful not to include the ID value to in serialized event Json.

agent-core/src/main/java/com/newrelic/agent/android/analytics/EventManagerImpl.java
```diff
@@ -363,7 +385,7 @@ public void onEventQueueSizeExceeded(int currentQueueSize) {
     /**
      * Notified when event buffer flush time has been exceeded. This normally occurs
      * between harvest cycles.
-     *
+     * <p>
```
cthomas-newrelic, line 388: Code cleanup replaces this in JavaDoc comments. Should be removed, but doesn't do any harm.

agent-core/src/test/java/com/newrelic/agent/android/analytics/AnalyticsEventTests.java
```diff
@@ -151,4 +151,10 @@ public void getMutableAttributeSet() {
             Assert.fail("Attributes should be mutable");
         }
     }
+
+    @Test
+    public void testEventUUID() {
+        AnalyticsEvent event = new AnalyticsEvent("testEvent");
+        Assert.assertNotNull(event.getEventUUID());
```
cthomas-newrelic, line 158: If it's really a UUID, you should test that it parses into a new UUID. Take a look at `UUID.fromString()`.

agent/src/main/java/com/newrelic/agent/android/stores/SharedPrefsEventStore.java
```diff
@@ -0,0 +1,85 @@
...
+        return false;
+    }
+
+    @Override
+    public List<AnalyticsEvent> fetchAll() {
+        final List<AnalyticsEvent> events = new ArrayList<AnalyticsEvent>();
+        for (Object object : super.fetchAll()) {
+            if (object instanceof String) {
+                try {
+                    JsonObject eventObj = new Gson().fromJson((String) object, JsonObject.class);
```
cthomas-newrelic, line 64: You need to replace the UUID (which is the key for each entry) after recreating the event from Json.  `newFromJson()` will create a new event, with a new uuid. `super.fetchAll())` doesn't provide the keys (UUIDS), so you'll have to use `sharedPrefs.getAll()` to return the data as a map<>, iterate through each entry and replace the UUID in the deserialize event with the key.

agent/src/main/java/com/newrelic/agent/android/stores/SharedPrefsEventStore.java
```diff
@@ -0,0 +1,85 @@
...
+        synchronized (this) {
+            try {
+                final JsonObject jsonObj = event.asJsonObject();
+                String eventJson = jsonObj.toString();
+
+                // events should be stored synchronously, since the app is terminating
+                SharedPreferences.Editor editor = this.sharedPrefs.edit();
+                editor.putString(event.getEventUUID(), eventJson);
+
+                StatsEngine.get().inc(MetricNames.SUPPORTABILITY_EVENT_SIZE_UNCOMPRESSED, eventJson.length());
```
cthomas-newrelic, line 49: For supportability metrics, use `StatsEngine.SUPPORTABILITY.inc(...)`as its decorator will also create metrics based on framework.

cthomas-newrelic (changes requested):
Almost there. Event deserialization needs to be fixed. It wouldn't matter until you tried to delete events, or some other problem occurred that prevented their upload.

cthomas-newrelic:
Please optimize imports before commits.

agent/src/main/java/com/newrelic/agent/android/stores/SharedPrefsEventStore.java
```diff
@@ -0,0 +1,85 @@
...
+    @Override
+    public List<AnalyticsEvent> fetchAll() {
+        final List<AnalyticsEvent> events = new ArrayList<AnalyticsEvent>();
+        for (Object object : super.fetchAll()) {
+            if (object instanceof String) {
+                try {
+                    JsonObject eventObj = new Gson().fromJson((String) object, JsonObject.class);
+                    events.add(AnalyticsEvent.newFromJson(eventObj));
+                } catch (Exception e) {
+                    log.error("Exception encountered while deserializing crash", e);
```
cthomas-newrelic, line 67: Dealing with an `event` here, not a `crash`.

agent/src/test/java/com/newrelic/agent/android/AndroidAgentImplTest.java
```diff
@@ -306,19 +309,20 @@ public void start() {
         Assert.assertNotNull("Should contain app background event", getSessionEvent(queuedEvents));
 
         agentImpl.applicationForegrounded(e);
-        assertEquals("Should contain foreground user action events", eventManager.getEventsRecorded(), 1);
+        assertEquals("Should contain foreground user action events", eventManager.getEventsRecorded(), 4);
```
cthomas-newrelic, line 312: Why would the counts here (and below) be higher? What are the additional events?

ywang-nr: Because when we persist the events, the total counts are growing while going foreground/background without sending payload: 
When DT is enabled, it changed from 1,3,1,3 --> 1,3,4,6
When DT is not enabled, it changed from 0,1,0,1 --> 0,1,1,2

But if we would like to simulate the similar scenario while app goes to background will send one last payload, I can clean the eventStore.

cthomas-newrelic: Yes, the test should reflect the actual lifecycle and state of the agent. With DT enabled, it shows the test is not resetting the event manager state when the agent restarts.

cthomas-newrelic (approved):
Please squash commit. 🚢 it

PR #149, opened 2023-11-03
fix: limit code coverage to PR open/reopen

cthomas-newrelic:
* adds codecov config to allow lower fail limits


ywang-nr (approved):
Added self-review in case you would like to merge it.

PR #152, opened 2023-11-27
[NR-121370] Add remote logging interface

cthomas-newrelic:
* adds log levels, level methods and stub `notice()` method.


ywang-nr (approved):
Great start for sure!

PR #155, opened 2023-11-29
Feature/nr121370 logging api

ywang-nr:
**WIP**
DONE:
1. Static API function names are settled
2. Auto instrumentation is added but not going live for phase 1

TODO: 
1. How are we printing out attributes and throwable with attributes
2. More unit tests


cthomas-newrelic (approved):
🐑 it

PR #166, opened 2024-01-12
Temp Solution to setEntityGuid

ywang-nr:
1. Temp solution for client side to setEntityGuid
2. Tested with apps, didn't write unit test as this will be removed later


ywang-nr:
Tested build with apps, didn't write unit test as this will be removed later.

agent-core/src/main/java/com/newrelic/agent/android/logging/LogReporting.java
```diff
@@ -25,6 +25,8 @@ public abstract class LogReporting {
     static LogLevel logLevel = LogLevel.INFO;
     static LogReporting instance = new LogReporting() {};
 
+    public static String entityGuid = "";
```
cthomas-newrelic, line 28: Does not need to be public if you provide gettor/settor.

cthomas-newrelic (approved):

PR #171, opened 2024-01-26
Feature/nr170090 log forwarder

cthomas-newrelic:


agent-core/src/main/java/com/newrelic/agent/android/logging/RemoteLogger.java
```diff
@@ -0,0 +1,343 @@
...
+                if (attributes != null) {
+                    /**
+                     * "attributes" Object: This sub-object contains all other attributes of the message
+                     *
+                     * Number of attributes per event: 255 maximum.
+                     * Length of attribute name: 255 characters.
+                     * Length of attribute value: 4,094 characters are stored in NRDB as a Log event field
+                     **/
+
+                    logDataMap.put(LogReporting.LOG_ATTRIBUTES_ATTRIBUTE, attributes);
```
ndesai-newrelic, line 164: why we are not iterating the attributes instead of appending with our log default attribute?

agent-core/src/main/java/com/newrelic/agent/android/logging/LogForwarder.java
```diff
@@ -0,0 +1,64 @@
...
+import com.newrelic.agent.android.util.Constants;
+
+import java.io.IOException;
+import java.net.HttpURLConnection;
+import java.net.URL;
+import java.util.concurrent.TimeUnit;
+
+import javax.net.ssl.HttpsURLConnection;
+
+public class LogForwarder extends PayloadSender {
```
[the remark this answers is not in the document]
cthomas-newrelic: Probably. now, I want to implement delivery using the Payload controller pattern used for Crash and Hex. It's a WIP, and if we don;t use it I'll clean up prior to GA

agent-core/src/main/java/com/newrelic/agent/android/logging/RemoteLogger.java
cthomas-newrelic: There aere two Json models supprted by Log ingest: flat (with all attributes on teh root), and common block (with attributes under their own block). I'd like to see which works better for us. Again, to be resolved by GA.

cthomas-newrelic: I was also curious if we might want to isolate customer provided attributes from those we decorate the log message .

ndesai-newrelic (approved):

PR #173, opened 2024-02-05
[NR-218026] Add Unreal as ApplicationFramework

ndesai-newrelic:


agent-core/src/main/java/com/newrelic/agent/android/ApplicationFramework.java
```diff
@@ -14,5 +14,8 @@ public enum ApplicationFramework {
     Xamarin,
     MAUI,
```
ywang-nr, line 16: Do we want to remove the empty lines?

agent-core/src/main/java/com/newrelic/agent/android/ApplicationFramework.java
```diff
@@ -14,5 +14,8 @@ public enum ApplicationFramework {
...
 
-    Unity
+    Unity,
+
```
ywang-nr, line 18: Do we want to remove the empty lines?

ndesai-newrelic: done.

ywang-nr (approved):
Looks good

PR #175, opened 2024-02-06
[NR-221368] added support for android dynamic feature

ndesai-newrelic:
Our Android agent currently cancels the build process when customers are using Android dynamic features. However, with the latest update, customers can now successfully build their app while utilizing Android dynamic features.


ywang-nr (dismissed):
Great! Good to know!

cthomas-newrelic (changes requested):
This change alone is insufficient. Needs tests (integration, functional or both , see the repo) and the test harness must be updated to reflect support for the new module type.

ndesai-newrelic:
> This change alone is insufficient. Needs tests (integration, functional or both , see the repo) and the test harness must be updated to reflect support for the new module type.

@cthomas-newrelic i added newrelic plugin on feature module and there is already test for checking dynamic feature. which other tests I can add for this?.

cthomas-newrelic (commented):
I think what you're trying to do is write a dynamic feature test app, which is fine, but perhaps in a separate project.

PR #177, opened 2024-02-08
[NR-178365] Enable Log forwarding to Log Ingest

cthomas-newrelic:
## [NR-178365] Add log forwarding class

LogForwarder is derived fromPayload sender, which is a component of the PayloadController model and understands Vortex response codes. Unlike a PayloadSender, it wraps a file as its data store.

  * Refactor PayloadSender to adapt a file-based payload
  * Adds tests

## [NR-223001] Add suggested supportability names to metrics namespace

Metric Root: "Supportability/AgentHealth/LogReporting" with the following sub-keys:

"Init" - LogReporting feature has started 
"UploadTime" - Duration of upload requests to ingest
"UploadTimeOut" - Count of failed upload attempts due to connect tmeout
"UploadThrottled" - Count of uploads not accepted by ingest due to API limits
"FailedUpload" - Count of upload attempts that fail for otehr reasons
"Expired" - Log data that is removed from use device due to expiration
"Removed/Rejected" - Count of log records that are rejects by the agent's LogReporter log validation
"Size/Uncompressed" - the size of the log data payload

## [NR-170090] Store log date files in the app cache directory

LogReports are stored in the apps cache directory, rather than the persistent files directory. The o/s _may_ remove the oldest files when storage runs low, offloading some of the maintenance work from the reporter, but potentially resulting in unreported log file deletions.

## [NR-170090] Experimental: set minimum size on file archives

** For MVP evaluation **: Prevents sending small amounts of data to log ingest every minute. Translates to fewer API calls, but does contradict the goal of getting data into NR asap. 

Threshold value to Vortex payload limit div 4.

### DIFF
--- a/agent-core/src/main/java/com/newrelic/agent/android/logging/LogForwarder.java
+++ b/agent-core/src/main/java/com/newrelic/agent/android/logging/LogForwarder.java
@@ -5,22 +5,89 @@
 
 package com.newrelic.agent.android.logging;
 
+import com.newrelic.agent.android.Agent;
 import com.newrelic.agent.android.AgentConfiguration;
+import com.newrelic.agent.android.metric.MetricNames;
+import com.newrelic.agent.android.payload.Payload;
 import com.newrelic.agent.android.payload.PayloadSender;
+import com.newrelic.agent.android.stats.StatsEngine;
 import com.newrelic.agent.android.util.Constants;
+import com.newrelic.agent.android.util.Streams;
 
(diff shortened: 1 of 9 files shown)


ndesai-newrelic (approved):
looks good to me.

cthomas-newrelic:
@ndesai-newrelic Still has a failing test that doesn't fail for me locally. Could you pull the branch and run the build (with check)?

PR #187, opened 2024-02-29
Update functionalTests.yml to disable JDK17 func test

cthomas-newrelic:
JDK Gradle TestKit blows out heap. This test should have been disabled until fully resolved. We were lucky it ran before, but the dynamic feature merge wasn't vetted for this test.


ywang-nr (approved):
Great!

Issue #191, opened 2024-03-25
ANRs being reported after enabling Offline Storage


astamato:
After enabling [this configuration](https://docs.newrelic.com/docs/mobile-monitoring/new-relic-mobile/mobile-sdk/configure-settings/#android-offline-storage) in our app we've started seeing reports on Firebase Crashlytics with this stack:

```
com.hubspot.android.api.newrelic.NewRelicInitializer.init
ANR triggered by slow IO operations
NewRelicInitializer.kt:40
``` 

```
java.io.FileDescriptor.sync (Native method)
android.os.FileUtils.sync (FileUtils.java:256)
android.app.SharedPreferencesImpl.writeToFile (SharedPreferencesImpl.java:807)
android.app.SharedPreferencesImpl.access$900 (SharedPreferencesImpl.java:59)
android.app.SharedPreferencesImpl$2.run (SharedPreferencesImpl.java:672)
android.app.SharedPreferencesImpl.enqueueDiskWrite (SharedPreferencesImpl.java:691)
android.app.SharedPreferencesImpl.access$100 (SharedPreferencesImpl.java:59)
android.app.SharedPreferencesImpl$EditorImpl.commit (SharedPreferencesImpl.java:604)
com.newrelic.agent.android.stores.SharedPrefsEventStore.store (SharedPrefsEventStore.java:48)
```

We were able to trace it back to [this line](https://github.com/newrelic/newrelic-android-agent/blob/25d330d582f0864db0bf46ffae29249053330974/agent/src/main/java/com/newrelic/agent/android/stores/SharedPrefsEventStore.java#L48).

Any help appreciated.

Issue #139, opened 2023-10-09
Excessive string concatination in debug message


odewdney:
https://github.com/newrelic/newrelic-android-agent/blob/fec4c538316b83c318ca9fdb0e9d77bc16294994/agent/src/main/java/com/newrelic/agent/android/NewRelic.java#L471C14-L471C14

Should this logging.debug call be wrapped in an if logging.level >= DEBUG

odewdney:
yes, hence the description of "excessive" - a large amount of string manipulation is done, only for it to typically be thrown away as the user is typically not logging at debug level, for a public method that could be called with high frequency. ( 1000s/sec )

i dont think an optimiser could skip the work, as the level value could change by another thread while it was doing the concatenations, and change the outcome.

ywang-nr:
Create a story in backlog to pickup [here](https://new-relic.atlassian.net/browse/NR-180972)

PR #193, opened 2024-04-02
fix: offline storage anr issue and excessive string concatenation

ndesai-newrelic:
#139 #191

### DIFF
--- a/agent/src/main/java/com/newrelic/agent/android/NewRelic.java
+++ b/agent/src/main/java/com/newrelic/agent/android/NewRelic.java
@@ -472,9 +472,12 @@ public final class NewRelic {
         StatsEngine.notice().inc(MetricNames.SUPPORTABILITY_API
                 .replace(MetricNames.TAG_NAME, "recordMetric"));
 
-        log.debug("NewRelic.recordMetric invoked for name " + name + ", category: " + category +
-                ", count: " + count + ", totalValue " + totalValue + ", exclusiveValue: " + exclusiveValue +
-                ", countUnit: " + countUnit + ", valueUnit: " + valueUnit);
+        if(log.getLevel() >= AgentLog.AUDIT) {
+            StringBuilder logString = new StringBuilder();
+            log.audit(logString.append("NewRelic.recordMetric invoked for name ").append(name).append(", category: ").append(category)
+                    .append(", count: ").append(count).append(", totalValue ").append(totalValue).append(", exclusiveValue: ").append(exclusiveValue)
+                    .append(", countUnit: ").append(countUnit).append(", valueUnit: ").append(valueUnit).toString());
+        }
         checkNull(category, "recordMetric: category must not be null. If no MetricCategory is applicable, use MetricCategory.NONE.");
 
         checkEmpty(name, "recordMetric: name must not be empty.");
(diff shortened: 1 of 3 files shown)


agent/src/main/java/com/newrelic/agent/android/NewRelic.java
```diff
@@ -472,9 +472,11 @@ public static void recordMetric(String name, String category, int count, double
         StatsEngine.notice().inc(MetricNames.SUPPORTABILITY_API
                 .replace(MetricNames.TAG_NAME, "recordMetric"));
 
-        log.debug("NewRelic.recordMetric invoked for name " + name + ", category: " + category +
-                ", count: " + count + ", totalValue " + totalValue + ", exclusiveValue: " + exclusiveValue +
-                ", countUnit: " + countUnit + ", valueUnit: " + valueUnit);
+        StringBuilder logString = new StringBuilder();
```
cthomas-newrelic, line 475: How does this reduce excessive string concatenation? IIRC, the complaint was the agent doing the work to build this excessive `DEBUG` level log message, even when logLevel was set to something below DEBUG. The change, imo, is to wrap the call in a conditional that checks the current log level first, before constructing the message.

ndesai-newrelic: String builder is more efficient when two or more strings are concat, i agree need to add log level before concatenation happen.

agent/src/main/java/com/newrelic/agent/android/stores/SharedPrefsEventStore.java
```diff
@@ -74,7 +74,7 @@ public void delete(AnalyticsEvent event) {
         try {
             synchronized (this) {
                 final SharedPreferences.Editor editor = sharedPrefs.edit();
-                editor.remove(event.getEventUUID()).commit();
+                editor.remove(event.getEventUUID()).apply();
```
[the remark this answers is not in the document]
ndesai-newrelic: this is created ANR if we are adding two many events on main thread.

agent/src/main/java/com/newrelic/agent/android/NewRelic.java
```diff
@@ -472,9 +472,12 @@ public static void recordMetric(String name, String category, int count, double
         StatsEngine.notice().inc(MetricNames.SUPPORTABILITY_API
                 .replace(MetricNames.TAG_NAME, "recordMetric"));
 
-        log.debug("NewRelic.recordMetric invoked for name " + name + ", category: " + category +
-                ", count: " + count + ", totalValue " + totalValue + ", exclusiveValue: " + exclusiveValue +
-                ", countUnit: " + countUnit + ", valueUnit: " + valueUnit);
+        if(log.getLevel() >= AgentLog.AUDIT) {
```
cthomas-newrelic, line 475: You replaced a `debug()` call with `AUDIT`. Is that intended?

cthomas-newrelic (approved):
🐑

agent/src/main/java/com/newrelic/agent/android/NewRelic.java
ndesai-newrelic: yes, you mentioned it here. https://github.com/newrelic/newrelic-android-agent/issues/139#issuecomment-1753256540

PR #194, opened 2024-04-02
[NR-59017] TicToc time calculation

ywang-nr:
Ticket: https://new-relic.atlassian.net/browse/NR-59017

What's changed:
1. Use System.nanoTime() rather than System.currentTimeMillis() to prevent time difference due to a network change
2. Add unit tests


ywang-nr:
Q: What was the reasoning for doing this?
A: Use nanoTime() will prevent the scenario that when time got changed on the device due to wifi network, which causes inconsistent tictoc time period. 

Q: What possible side effects might we see due to using System.currentTimeMillis() elsewhere in the agent?
A: I double checked most of other places of using System.currentTimeMills(), most of them are used as a timestamp. We could change few places that use to calculate time difference, but that will be a different request than this PR.

agent-core/src/main/java/com/newrelic/agent/android/stats/TicToc.java
```diff
@@ -17,11 +17,11 @@ private static enum State {
 
     public void tic() {
```
cthomas-newrelic, line 18: Could you please return the TicToc instance here? ```public TicTok tic() { return this; }```

ywang-nr: Fixed

agent-core/src/main/java/com/newrelic/agent/android/stats/TicToc.java
```diff
@@ -32,7 +33,31 @@ public long toc() {
     }
 
     public long peek() {
-        return (state == State.STARTED) ? System.currentTimeMillis() - startTime : 0;
+        return (state == State.STARTED) ? System.nanoTime() - startTime : 0;
+    }
+
+    public long getStartTime() {
```
cthomas-newrelic, line 39: Do we need accessors to set start/end time? The class encapsulates a time span, so I wouldn't think either value would be manually changed by the caller.

ywang-nr: I mainly use the getter/setter to verify unit tests.

cthomas-newrelic: Give the fields and methods package private or protected access, and tests can access them.

PR #205, opened 2024-04-26
TicToc time calculation & convertion

ywang-nr:
Ticket: https://new-relic.atlassian.net/browse/NR-59017

What's changed:
1. Use nanoTime() but covert to millionSeconds due to agent is using ms as default calculation.
2. Add unit tests


cthomas-newrelic (approved):
:shipit:

PR #209, opened 2024-05-01
[NR-247776] Report AEI all reasons

ywang-nr:
What's changed:
1. We would like to report all reasons rather than just ANR reason
2. Modify existing unit tests


cthomas-newrelic (approved):
👍

cthomas-newrelic:
Oops, just noticed this PR was a draft.

ywang-nr:
> Oops, just noticed this PR was a draft.

Yeah, I was looking into whether we can add more unit tests for other reasons.

cthomas-newrelic (commented):
I think this could benefit from additional tests, such as verification that the AEI records we parse end up as correct MobileApplicationExit events with accompanying attributes. Look at the tests I added regarding the importance field in `testApplicationExitImportance`.

PR #213, opened 2024-05-07
fix: Newrelic Agent is not starting when app comes in foreground from…

ndesai-newrelic:
… the background

Maui agent is initiating the applicationMontinor activity from multiple locations, causing the activity count values to increase multiple times. Additionally, when the app transitions to the background, the count is not reset to 0. Consequently, the agent fails to restart when the app returns to the foreground.


cthomas-newrelic (changes requested):
Looks good, tests please. We do have ApplicationStateMonitor test in place.

ndesai-newrelic:
this change is not required.

PR #225, opened 2024-06-04
fix: The handling of event storage in shared preferences is not being…

ndesai-newrelic:
… done correctly, leading to Application Not Responding (ANR) errors and a long delay in saving the events.


agent/src/main/java/com/newrelic/agent/android/stores/SharedPrefsEventStore.java
```diff
@@ -45,7 +43,8 @@ public boolean store(AnalyticsEvent event) {
                 editor.putString(event.getEventUUID(), eventJson);
 
                 StatsEngine.SUPPORTABILITY.inc(MetricNames.SUPPORTABILITY_EVENT_SIZE_UNCOMPRESSED, eventJson.length());
-                return editor.commit();
+                editor.apply();
```
cthomas-newrelic, line 46: This should be calling applyOrCommitEditor(), but I just noticed that method has been changed as well. There should be a commit() as a fail safe if apply throws an exception.

cthomas-newrelic (changes requested):
The commit message needs fixing. Try to adhere to the [50/72 rule](https://gist.github.com/robertpainsi/b632364184e70900af4ab688decf6f53) when formatting commi t messages.

agent/src/main/java/com/newrelic/agent/android/stores/SharedPrefsEventStore.java
ndesai-newrelic: i updated it.

PR #228, opened 2024-06-05
setUserId()/Session times discussion

cthomas-newrelic:
# DO NOT MERGE

### DIFF
--- a/agent-core/src/main/java/com/newrelic/agent/android/analytics/AnalyticsAttribute.java
+++ b/agent-core/src/main/java/com/newrelic/agent/android/analytics/AnalyticsAttribute.java
@@ -186,30 +186,36 @@ public class AnalyticsAttribute {
         return (attributeDataType == AttributeDataType.STRING) ? stringValue : null;
     }
 
-    public void setStringValue(String stringValue) {
+    public AnalyticsAttribute setStringValue(String stringValue) {
         this.doubleValue = Double.NaN;
         this.stringValue = stringValue;
         this.attributeDataType = AttributeDataType.STRING;
+
+        return this;
     }
 
     public double getDoubleValue() {
         return (attributeDataType == AttributeDataType.DOUBLE) ? doubleValue : Double.NaN;
     }
 
-    public void setDoubleValue(double doubleValue) {
+    public AnalyticsAttribute setDoubleValue(double doubleValue) {
         this.doubleValue = doubleValue;
         this.stringValue = null;
         this.attributeDataType = AttributeDataType.DOUBLE;
+
+        return this;
     }
 
     public boolean getBooleanValue() {
         return (attributeDataType == AttributeDataType.BOOLEAN) ? Boolean.valueOf(stringValue).booleanValue() : false;
     }
 
-    public void setBooleanValue(boolean boolValue) {
+    public AnalyticsAttribute setBooleanValue(boolean boolValue) {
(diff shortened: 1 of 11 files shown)


agent-core/src/main/java/com/newrelic/agent/android/analytics/AnalyticsAttribute.java
```diff
@@ -186,39 +186,47 @@ public String getStringValue() {
         return (attributeDataType == AttributeDataType.STRING) ? stringValue : null;
     }
 
-    public void setStringValue(String stringValue) {
+    public AnalyticsAttribute setStringValue(String stringValue) {
```
cthomas-newrelic, line 189: This an similar changes below will be reverted.

agent-core/src/main/java/com/newrelic/agent/android/harvest/Harvest.java
```diff
@@ -73,21 +73,20 @@ public static void stop() {
...
     }
 
-    public static void harvestNow(boolean finalizeSession) {
-        if (!isInitialized())
-            return;
-
-        if (finalizeSession) {
-            // add session meta to next harvest
-            instance.finalizeSession();
+    public static void harvestNow(boolean finalizeSession, boolean bWait) {
```
cthomas-newrelic, line 76: This new method encapsulated changes to HarvestTimer to _restart the session_, which is to say resetting the the harvest times. Session times (total, time since start) are derived from this clock. It has not been hooked up to exiting code.

agent-core/src/main/java/com/newrelic/agent/android/harvest/HarvestTimer.java
```diff
@@ -145,20 +140,46 @@ public void shutdown() {
...
-    // Runs a tick of the Harvester immediately, disregarding any 'time since last tick' limits.
-    public void tickNow() {
-        final HarvestTimer timer = this;
-        ScheduledFuture future = scheduler.schedule(new Runnable() {
-            @Override
-            public void run() {
-                timer.tick();
-            }
-        }, 0, TimeUnit.SECONDS);
+    public void restart() {
```
cthomas-newrelic, line 143: This new method encapsulated changes to HarvestTimer to _restart the session_, which is to say resetting the the harvest times. Session times (total, time since start) are derived from this clock. It has not been hooked up to exiting code.

agent/src/main/java/com/newrelic/agent/android/NewRelic.java
```diff
@@ -927,13 +927,29 @@ public static boolean removeAllAttributes() {
...
+        final AnalyticsControllerImpl controller = AnalyticsControllerImpl.getInstance();
+        final AnalyticsAttribute userIdAttr = controller.getAttribute(AnalyticsAttribute.USER_ID_ATTRIBUTE);
+
+        if (userIdAttr != null) {
+            if (userId == null || userId.isEmpty()) {
+                controller.removeAttribute(AnalyticsAttribute.USER_ID_ATTRIBUTE);
+            }
+
+            if (userIdAttr.getStringValue() != userId) {
+                Harvest.harvestNow(true, false);   // call non-blocking harvest
```
cthomas-newrelic, line 945: Blocking methods should never be called from the NewRelic API level. The new 2nd arg says whether to wait on harvest completion, which is no in  this case.

PR #264, opened 2024-08-14
fix: Agent Version and App Build Id Mismatch After App Update

ndesai-newrelic:
Agent is sending newer data token with Crash Created in Older version of the app.

### DIFF
--- a/agent-core/src/main/java/com/newrelic/agent/android/crash/CrashReporter.java
+++ b/agent-core/src/main/java/com/newrelic/agent/android/crash/CrashReporter.java
@@ -114,11 +114,11 @@ public class CrashReporter extends PayloadReporter {
     }
 
     protected Future reportCrash(final Crash crash) {
-        final boolean hasValidDataToken = Harvest.getHarvestConfiguration().getDataToken().isValid();
+        if (crash != null) {
+            final boolean hasValidDataToken = crash.getDataToken().isValid();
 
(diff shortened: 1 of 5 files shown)


agent-core/src/main/java/com/newrelic/agent/android/crash/CrashReporter.java
```diff
@@ -114,11 +114,11 @@ protected void reportSavedCrashes() {
     }
 
     protected Future reportCrash(final Crash crash) {
-        final boolean hasValidDataToken = Harvest.getHarvestConfiguration().getDataToken().isValid();
+        if (crash != null) {
+        final boolean hasValidDataToken = crash.getDataToken().isValid();
```
cthomas-newrelic, line 118: The formatting is off here

agent-core/src/test/java/com/newrelic/agent/android/test/mock/Providers.java
```diff
@@ -296,6 +310,7 @@ public static HarvestConfiguration provideHarvestConfiguration() {
         return harvestConfiguration;
     }
 
+
```
cthomas-newrelic, line 313: Excessive newline.

cthomas-newrelic (dismissed):
A few minor formatting things, but looks good to me.

ywang-nr (approved):

PR #299, opened 2024-10-24
[NR-331312] Add AEI session mapper

cthomas-newrelic:
Acceptance criteria:

* a PID-to-session map will be created and persisted in local AEI cache
* for AEI events gathered during the first agent run, the sessionID of the event will be the one that recorded the event
* for AEI events gathered in successive sessions, a new attribute will be added to provide the session ID in which the AEI occurred 
  * in this case session ID will never match recorded session ID
* session ID map will be updated each time the agent starts
* session ID map will be updated each time the agent restarts
* session ID map will be pruned when the set of AEI records returned by the ART no longer includes elements in the map
* an attribute with the historic session ID will be inserted into the event as `sessionId`.

Tests will be created to reflect this criteria.

### DIFF
--- a/agent-core/src/main/java/com/newrelic/agent/android/aei/SessionMapper.java
+++ b/agent-core/src/main/java/com/newrelic/agent/android/aei/SessionMapper.java
@@ -51,6 +51,7 @@ public class SessionMapper {
         return sessionId.isEmpty() ? defaultSessionId : sessionId;
     }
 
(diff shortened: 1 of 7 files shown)


cthomas-newrelic:
Shoot, this test fix is in another PR.  I can port it to this branch, or wait until the other PR is merged.

ndesai-newrelic (approved):

PR #300, opened 2024-10-25
[NR-315570] Add AEI trace model

cthomas-newrelic:
* parses system trace provided by AEI record into thread collection
* adds tests


cthomas-newrelic:
Shoot, this test fix is in another PR. I fixed it in this branch as well.

ndesai-newrelic (approved):

Issue #261, opened 2024-07-31
Add ability to disable default log instrumentation through gradle


ttsatsanis:
## Summary

Nowadays, most applications do not depend on the default Android Log implementation. Instead they use SFL4J, Logback or other frameworks.
While the ability to upload logs is a great advantage, one cannot upload its specific logs in their desired format.

## Desired Behavior

Have the ability to disable log instrumentation through new relic gradle config and be able to use my own `NewRelic.logxxx` functions.

## Possible Solution

Maybe something like the following would be enough, allowing the gradle plugin to not run the log instrumentation (not an actual possible solution):

```
newRelic {
     ...
     setLogInstrumentationEnabled(boolean: true|false)
}
```

## Additional context

Since NewRelic gives the ability to dynamically create columns and distribute data in each column, users must have the ability to enhance their search criteria based on their logging needs, with the support of their desired logging facility.

ndesai-newrelic:
@ttsatsanis We'd like to understand your specific use case for disabling our logging instrumentation. Our solution currently instruments all major logging libraries, providing comprehensive coverage. Could you elaborate on the reasons behind your decision to turn off this feature? We're particularly interested in learning about any scenarios or requirements that our current logging instrumentation doesn't address.
This information will help us better understand your needs and potentially improve our offering. Are there particular aspects of our logging instrumentation that don't align with your requirements?

sguptaruby:
Same question from my side how do I disable default log, I want track those loges which I have sent with the help of this function NewRelic.logAttributes.

ttsatsanis:
@ndesai-newrelic sorry for the delayed response!!

I found out (which was obvious, but did not understood well) that I have been using a LogcatAppender for SLF4J and that's where NewRelic instrumented these functions and send logs to the platform. After not applying this, I got pure logs from `NewRelic.logAttributes` and not duplicate logs from instrumented Logcat and my logging implementation.
However, although my issue is solved, I think that the possible solution I proposed can allow a developer to maintain logcat logs (`Log.x`) and also apply its own logging implementation with the use of `NewRelic.logxxxx` without the need to applying or not applying the Logcat interface.

Let me explain this through code.

Here is a logging implementation with SLF4J and Logback Android:

```kotlin
private fun installLogger(
        enableLogReporting: Boolean,
        logLevel: Level = Level.INFO
    ) {
        val context = LoggerFactory.getILoggerFactory() as LoggerContext
        context.stop()

[...]

PR #330, opened 2024-12-17
[NR-329587] feat: added ability to disable logInstrumentation at Buil…

ndesai-newrelic:
…d  Time

This feature eliminates the mappings for log replacement calls when the `logInstrumentationEnabled` flag is set to `false` at build time. By default, this flag is set to `true`.

#261


smalsam-newr (dismissed):

instrumentation/src/main/java/com/newrelic/agent/InstrumentationAgent.java
```diff
@@ -157,7 +157,8 @@ private static void createInvocationDispatcher(org.slf4j.Logger log) throws Exce
         if (field.get(null) instanceof InvocationDispatcher) {
             log.info("Detected cached instrumentation.");
         } else {
-            field.set(null, new InvocationDispatcher(log));
+            boolean logInstrumentationEnabled = agentOptions.get("logInstrumentationEnabled") != null && agentOptions.get("logInstrumentationEnabled").equals("true");
```
cdillard-NewRelic, line 160: Shall string "logInstrumentationEnabled" be a constant?

cdillard-NewRelic (dismissed):
Approved w one q about string constant and wondering if its OK to enable by default.
I think logInstrumentationEnabled should probably be disabled by default like on iOS but your choice. 
On iOS we've implemented this feature flag as disabled by default and named "NRFeatureFlag_AutoCollectLogs"

instrumentation/src/main/java/com/newrelic/agent/InstrumentationAgent.java
ndesai-newrelic: it should be, I will add this as cosntant.

ndesai-newrelic:
@cdillard-NewRelic it was enabled by default when we  released logs, we are now giving them capability if they want to disable. I will check with PM for this.

agent-core/src/test/java/com/newrelic/agent/android/logging/RemoteLoggerTest.java
```diff
@@ -183,6 +183,36 @@ public void checkCommonAttributes() throws Exception {
...
+        Assert.assertTrue(attributes.has("osName"));
+        Assert.assertTrue(attributes.has("platformVersion"));
+        Assert.assertTrue(attributes.has("newRelicVersion"));
+        Assert.assertTrue(attributes.has("deviceModel"));
+        Assert.assertTrue(attributes.has("deviceModel"));
+        Assert.assertTrue(attributes.has("deviceManufacturer"));
+    }
+
+    @Test
+    public void checkSessionAttributesFromCommonAttributes() throws Exception {
```
cdillard-NewRelic, line 197: Nice

cdillard-NewRelic (approved):
👍

PR #347, opened 2025-03-19
fix:fixed compile errors for DexGuard [NR-375993]

ndesai-newrelic:


cdillard-NewRelic (approved):
Looks good for a DexGuard syntax update.

PR #348, opened 2025-03-21
fix:[NR-382146] Android Agent Fails to Parse Non-Zero Float/Double Values for log Sampling Rate

ndesai-newrelic:


smalsam-newr (approved):

cdillard-NewRelic (approved):
Lookin good

PR #349, opened 2025-03-25
fix:sanitize the Session/Start Metrics before sending to server

ndesai-newrelic:


smalsam-newr (approved):

cdillard-NewRelic (approved):
Approved the sanitization.

PR #351, opened 2025-03-26
Release/7.6.5

ndesai-newrelic:


agent/src/main/java/com/newrelic/agent/android/util/ComposeChecker.java
```diff
@@ -0,0 +1,28 @@
+package com.newrelic.agent.android.util;
+
+import android.content.Context;
+
+import java.util.Arrays;
+import java.util.List;
+
+public class ComposeChecker {
```
cdillard-NewRelic, line 8: This is cool. Excited to add something similar for SwftUI detection in iOS.

cdillard-NewRelic (approved):
👍 approved the release

PR #365, opened 2025-05-08
NR-409074 Log ANRs Fix

ywang-nr:
Ticket: [NR-409074](https://new-relic.atlassian.net/browse/NR-409074)

What's changed:
1. Log reporting files are backed up instead of deleting, could be the potential cause of ANRs, changing it to a deletion process.
2. We are only using two files at time to write into data and upload to the server.
3. Remove the size checking of 1MB because we are using gzip when uploading.
4. Added cleanup process onHarvestStart
5. Truncated messages issues: partially fixed, still need to fix the race condition issue on every harvest cycle with background reporting on/off.
6. Warning message: resources may not be closed? - this is our logwriter, we need to leave it open for log writing but will see how we can improve it.

[NR-409074]: https://new-relic.atlassian.net/browse/NR-409074?atlOrigin=eyJpIjoiNWRkNTljNzYxNjVmNDY3MDlhMDU5Y2ZhYzA5YTRkZjUiLCJwIjoiZ2l0aHViLWNvbS1KU1cifQ

### DIFF
--- a/agent-core/src/main/java/com/newrelic/agent/android/harvest/Harvester.java
+++ b/agent-core/src/main/java/com/newrelic/agent/android/harvest/Harvester.java
@@ -14,7 +14,6 @@ import com.newrelic.agent.android.FeatureFlag;
 import com.newrelic.agent.android.TaskQueue;
 import com.newrelic.agent.android.activity.config.ActivityTraceConfiguration;
 import com.newrelic.agent.android.activity.config.ActivityTraceConfigurationDeserializer;
-import com.newrelic.agent.android.analytics.AnalyticsAttribute;
 import com.newrelic.agent.android.analytics.AnalyticsControllerImpl;
 import com.newrelic.agent.android.background.ApplicationStateMonitor;
 import com.newrelic.agent.android.logging.AgentLog;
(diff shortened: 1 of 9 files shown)


ywang-nr:
Sqash Merge!!!

ywang-nr:
> @ywang-nr Hi, I've tested the agent test build, and there is some improvement. After a fresh start of the app, only the first two logs are missing. Previously, the entire first batch (even thousands of logs) was missing. Now, the logs are uploaded reliably at the beginning.
> 
> However, once I put my app into the background - for example, by pressing the Home button - the following logs are printed:
> 
> ```
> UI has become hidden (app backgrounded)
> AndroidAgentImpl: application backgrounded
> ```
> 
> Then the agent starts to break. After reopening the test app, I see:
> 
> ```
> A resource failed to call close. 
> ```
> 
> and then the error occurs:
> 
> ```
> Invalid Json entry skipped [{"level":"INFO","attributes]
> ```
> 
> This error again causes the logs to go missing.
> 
> It looks like moving the test app into the background is triggering serious problems with New Relic agent. I wouldn't be surprised if two or more New Relic agent threads are starting to process the same data and triggering race condition errors.

Thanks for testing. I will take a look at the issue when app goes to background for sure. 
Just to confirm: 
1. Are you still experiencing ANR or OOM issue with this build?
2. We have noticed missing of the few initial lines of logs and has created a ticket for that for discussion, unfortunately it's because of the delay that we are waiting for the configurations to send back from the service side then start up the log reporter.

gregd-flatpay:
> Are you still experiencing ANR or OOM issue with this build?

@ywang-nr To be honest, I wasn't testing the new Agent test build in terms of memory usage and performance. However, I doubt there has been any improvement when I still see the `A resource failed to call close.` warnings.

ndesai-newrelic (approved):

PR #373, opened 2025-06-12
NR-426117_Bug_fix_Android

ndesai-newrelic:


ywang-nr (approved):
You might want to format each line a bit, otherwise looks good.

PR #393, opened 2025-08-12
[NR-438547] RecordMetric Investigate

ywang-nr:
Ticket: https://new-relic.atlassian.net/browse/NR-438547

**What's changed:**
1. See ticket for more detailed background and investigation information.
2. Customer cannot provide a reproduce case, neither we can reproduce it. 
3. Debug through the process, clean up the code to make it more straightforward, and then we will monitor it.

### DIFF
--- a/agent-core/src/main/java/com/newrelic/agent/android/measurement/CustomMetricMeasurement.java
+++ b/agent-core/src/main/java/com/newrelic/agent/android/measurement/CustomMetricMeasurement.java
@@ -19,9 +19,12 @@ public class CustomMetricMeasurement extends CategorizedMeasurement {
         setName(name);
         customMetric = new Metric(name);
 
-        // Call sample so that min, max and sos get updated. Then reset count to the provided value.
-        customMetric.sample(totalValue);
+        // set customMetric value one by one
         customMetric.setCount(count);
+        customMetric.setTotal(totalValue);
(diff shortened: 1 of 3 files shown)


agent-core/src/main/java/com/newrelic/agent/android/measurement/CustomMetricMeasurement.java
```diff
@@ -19,9 +19,12 @@ public CustomMetricMeasurement(String name, int count, double totalValue, double
         setName(name);
```
cmcadams-newrelic, line 19: Should we rename the function. By renaming this function to `SetCustomMetricMeasurement` Which highlights by conversion this function will only set the values for a `custom  metric`. I think this name change would also allow us to eliminate the comment on line 22.

cmcadams-newrelic (approved):
Nice work! :shipit: 

I left a non blocking comment. Feel free to ignore my comment or make the change before merging this PR.

agent-core/src/main/java/com/newrelic/agent/android/measurement/CustomMetricMeasurement.java
ywang-nr: This would be a constructor so we will leave it the same as the class name.

PR #402, opened 2025-09-09
NR-426877 fixes the trace listener issue which is not removed correctly.

ndesai-newrelic:


ywang-nr (approved):
LGTM, only comment is the agent version

PR #403, opened 2025-09-09
[NR-300843]Fix: OkHttp3 Url updated by interceptors

ndesai-newrelic:
When OkHttp3 interceptors modify the request URL, the New Relic Android agent now correctly reports the final, modified URL. This is achieved by adding a New Relic interceptor after all existing client interceptors. This interceptor captures the final URL from the request chain and updates the transaction state accordingly.


hoangnguyen92dn:
@ndesai-newrelic @ywang-nr Hi. Thanks for the fixing, the issue seems what we are facing when seeing NR logs always report original base URL before we change the path by using interceptor. I am checking to eliminate current workaround solution in our app. Curiously this fix is release on `7.6.9` but it's not visible in `7.6.10`? 

image

PR #404, opened 2025-09-09
[NR-458555] Telemetry data transmission from the Android Agent stops …

ndesai-newrelic:
…after transitioning to a specific screen.


ndesai-newrelic:
@ywang-nr I removed the the renterlock which is not fixed the issue earlier issue and created other issue related data forwarding.

PR #457, opened 2025-12-05
[NR-488394]Fix(dexguard): Correct DexGuard 9 mapping file path constr…

ndesai-newrelic:
…uction

Constructs the mapping file path for DexGuard 9 using `flavorName` and `buildType` instead of the full variant name. This resolves an issue where the incorrect path was generated for flavored builds.

Also adds extensive logging to the map upload task for improved debugging of execution conditions.


cdillard-NewRelic (approved):
Code seems okay but I'll defer to android dev for final say.

PR #475, opened 2026-03-17
fix: serialize data-nr-masked as top-level rrweb attribute

jgibson02:
## [NR-529779](https://new-relic.atlassian.net/browse/NR-529779)

fix: serialize data-nr-masked as top-level rrweb attribute

## What & Why

The `data-nr-masked` attribute (added in NR-533692) was being placed inside the rrweb
node's `style` object via the metadata map. During replay, rrweb applies style properties
via `element.style.setProperty()`, and the browser silently drops `data-nr-masked` as an
invalid CSS property — making it invisible to the player plugin.

This fix adds a dedicated `dataNrMasked` field on `Attributes` and serializes it as a
top-level attribute in `AttributesSerializer`, so rrweb applies it via `setAttribute()`
and the player plugin can detect it with `element.getAttribute('data-nr-masked')`.

## Callouts

- Affects both `SessionReplayImageViewThingy` (traditional views) and `ComposeImageThingy` (Compose)
- Both full snapshot and incremental mutation paths are updated

## Test plan

- Verified via rrweb-live-streaming that `data-nr-masked` appears as a top-level attribute
  in the rrweb JSON (not nested inside `style`)
- Verified on-device replay data files contain `"data-nr-masked":"image"` at the attribute level
- Player plugin successfully detects and replaces masked images in NR1 staging

[NR-529779]: https://new-relic.atlassian.net/browse/NR-529779?atlOrigin=eyJpIjoiNWRkNTljNzYxNjVmNDY3MDlhMDU5Y2ZhYzA5YTRkZjUiLCJwIjoiZ2l0aHViLWNvbS1KU1cifQ

### DIFF
--- a/agent/src/main/java/com/newrelic/agent/android/sessionReplay/SessionReplayImageViewThingy.java
+++ b/agent/src/main/java/com/newrelic/agent/android/sessionReplay/SessionReplayImageViewThingy.java
@@ -281,7 +281,7 @@ public class SessionReplayImageViewThingy implements SessionReplayViewThingyInte
     public RRWebElementNode generateRRWebNode() {
         Attributes attributes = new Attributes(viewDetails.getCssSelector());
         if (isMasked) {
-            attributes.getMetadata().put("data-nr-masked", "image");
+            attributes.dataNrMasked = "image";
         }
         return new RRWebElementNode(attributes, RRWebElementNode.TAG_TYPE_DIV, viewDetails.getViewId(), new ArrayList<>());
     }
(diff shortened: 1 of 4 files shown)


ndesai-newrelic:
@jgibson02 it is already implemented, we need changes from RRWeb Plater side to test these changes. https://github.com/newrelic/newrelic-android-agent/commit/bcfd93272add69ce2edeed8b2b110a0b4eeee880

jgibson02:
@ndesai-newrelic this PR is to address that the `data-nr-masked` attribute was being serialized inside the `styles` block instead of being on the DOM node itself

PR #485, opened 2026-04-03
[NR-527657] auto create new session at 4-hour mark

ywang-nr:
Ticket: https://new-relic.atlassian.net/browse/NR-527657

**What's changed:**
1. Auto create new session at 4-hour mark when app is running
2. Add unit tests


agent-core/src/main/java/com/newrelic/agent/android/harvest/HarvestTimer.java
```diff
@@ -22,6 +26,7 @@
 public class HarvestTimer implements Runnable {
     public final static long DEFAULT_HARVEST_PERIOD = TimeUnit.SECONDS.toMillis(60);
     private final static long HARVEST_PERIOD_LEEWAY = TimeUnit.SECONDS.toMillis(1);
+    private final static long SESSION_REPLAY_SESSION_PERIOD = TimeUnit.HOURS.toMillis(4);
```
ndesai-newrelic: why are you using session replay time period? this is for generic session

agent-core/src/main/java/com/newrelic/agent/android/harvest/Harvest.java
```diff
@@ -408,7 +408,7 @@ protected ActivityTraceConfiguration getActivityTraceConfiguration() {
         return harvestConfiguration.getAt_capture();
     }
 
-    void finalizeSession() {
+    public void finalizeSession() {
```
ndesai-newrelic: instead of changing in harvesttimer, why cant we start new session on harvest Class?

agent-core/src/main/java/com/newrelic/agent/android/harvest/Harvest.java
```diff
@@ -54,6 +55,7 @@ public void initializeHarvester(AgentConfiguration agentConfiguration) {
         harvester.setAgentConfiguration(agentConfiguration);
         harvester.setHarvestConfiguration(instance.getConfiguration());
         flushHarvestableCaches();
+        agentConfig = agentConfiguration;
```
ndesai-newrelic, line 58: where do we need agent configuration?

agent-core/src/main/java/com/newrelic/agent/android/harvest/Harvester.java
```diff
@@ -820,4 +822,23 @@ Collection<HarvestLifecycleAware> getHarvestListeners() {
         return new ArrayList<>(harvestListeners);
     }
 
+    /**
+     * Automatic session termination for the New Relic agent after 4 hours of continuous session time.
+     */
+    public void checkAndResetSessionIfExpired() {
+        try {
+            HarvestTimer harvestTimer = Harvest.getInstance().getHarvestTimer();
+            if (harvestTimer.sessionTimeSinceStart() >= HarvestTimer.DEFAULT_SESSION_DURATION_PERIOD) {
```
ndesai-newrelic, line 831: can we use getMillisSinceStart() instead of this new method?

ndesai-newrelic (changes requested):

agent-core/src/main/java/com/newrelic/agent/android/harvest/Harvest.java
ywang-nr: agentConfig.provideSessionId()

agent-core/src/main/java/com/newrelic/agent/android/harvest/Harvester.java
ywang-nr: I changed it, but want to leave the new function for now until testing is all set.

ndesai-newrelic:
@ywang-nr can you check this one? 

  1. checkAndResetSessionIfExpired uses wrong timer — will reset every 60s after 4 hours

  Harvester.java:184 calls Harvest.getMillisSinceStart(), which reads harvestTimer.timeSinceStart() (based on startTimeMs — the original app start). But startSession() at Harvest.java:34 only resets
  sessionStartTimeMs, not startTimeMs.

  After the first 4-hour reset, getMillisSinceStart() still returns >4h, so checkAndResetSessionIfExpired() will fire every single harvest cycle (every 60 seconds) forever. The session ID will churn on
  every harvest.

  2. agentConfig can be null — NPE in startSession

  Harvest.java:41 — agentConfig is set in initializeHarvester(), but startSession() at line 41 calls agentConfig.provideSessionId() without a null check. If startSession is called before
  initializeHarvester (e.g., via a test or race condition), this NPEs. The try-catch will swallow it silently.


  3. finalizeSession visibility change: void → public void

  Harvest.java:51 — finalizeSession() was package-private, now it's public. This expands the API surface. If this is intentional for Harvester to call it, consider whether protected would be more
  appropriate.

  4. startSession doesn't finalize the old session

  Harvest.startSession() creates a new session ID and removes SESSION_DURATION_ATTRIBUTE, but never calls finalizeSession() to flush the old session's data (session event, duration metric). The previous
  session's analytics will be lost — no SessionEvent is recorded for the 4-hour session that just ended.


  5. Duplicate log messages in startSession

  Harvest.java:35-37:
  log.debug("Harvest: Session replay limit reached (4 hours). Resetting session start time.");
  log.debug("Harvest: Session replay limit reached (4 hours). Creating new session.");

  Two nearly identical log lines in succession. Also, these messages mention "Session replay limit" but this is a general session reset — the method is called startSession, not something replay-specific.

ywang-nr:
1. This is the reason I need to use the new function, don't want to change the old timer as it's used somewhere else in the agent.
2. Added null check
3. Changed to protected
4. finalizeSession() is called inside harvestNow() before startSession(), no change
5. refined the messages

ndesai-newrelic (approved):

Issue #495, opened 2026-04-24
NewRelicConfigTask invalidates Gradle build cache on every build


trevjonez:
## Summary

The `NewRelicConfigTask` generates a `NewRelicConfig.java` source file containing a randomly generated `BUILD_ID` (via `UUID.randomUUID()`) and a build metrics string that change on every clean build. This generated source file is wired into the consuming app's compile task inputs via `addGeneratedSourceDirectory()` (AGP 7.4+) or `registerJavaGeneratingTask()` (AGP 4.x). Because the source content changes every build, the Java/Kotlin compile tasks see different inputs and cannot use Gradle's build cache — local or remote. Every CI build performs a full recompilation of the consuming app, even when no source code has actually changed.

## Desired Behavior

Dynamic values generated by the New Relic Gradle plugin (build ID, metrics) should not invalidate the consuming app's compile or dex tasks. Expensive, cacheable tasks like `compileReleaseJavaWithJavac`, `compileReleaseKotlin`, and `minifyReleaseWithR8` should remain cache-hits across CI builds when only the build ID changes.

## Possible Solution

Replace the generated `NewRelicConfig.java` source file with a `newrelic_config.json` file placed in Android assets via `addGeneratedSourceDirectory()` on `variant.sources.assets`. At runtime, read the JSON from `context.getAssets()` during agent initialization and cache the values.

**Why assets:**
- Assets are not compiled or dexed. They are merged during `mergeAssets` and packaged during APK assembly — both fast, I/O-bound tasks that are never meaningfully cached.
- Compile tasks (`javac`, `kotlinc`) and dex tasks (R8/D8) never see the dynamic values, so they remain fully cacheable.
- The build ID stays as a random UUID — no correctness risk, no change to the ID generation strategy.
- DexGuard compatibility is preserved — DexGuard's asset encryption is transparent at runtime via its hooked `AssetManager`.

**Alternatives considered and rejected:**

1. **Make the build ID deterministic** (e.g., derive from git SHA). Rejected because the build ID correlates crash reports with ProGuard/R8 mapping files. A deterministic ID would need to account for every possible input that affects the mapping file to guarantee correctness. This is extremely difficult and a failure (wrong mapping applied to a crash) is worse than a build performance issue.

2. **Generate bytecode directly via ASM** and inject into the class pipeline after compilation. Rejected because the generated class JAR still feeds into R8/D8 dexing. This shifts the cache invalidation from compilation to dexing rather than eliminating it.

[...]

PR #496, opened 2026-04-24
[NR-556102] Fix NewRelicConfigTask invalidating Gradle build cache on every build

trevjonez:
Fixes https://github.com/newrelic/newrelic-android-agent/issues/495

## Summary

- Replace generated `NewRelicConfig.java` source file with a `newrelic_config.json` asset file
- Plugin generates JSON via `JsonOutput` and wires it as a generated asset directory
- Runtime reads config from assets during `AndroidAgentImpl` initialization
- `Agent.getBuildId()` and `Agent.getIsObfuscated()` use cached fields instead of reflection

## Motivation

The plugin generates `NewRelicConfig.java` with a random `BUILD_ID` that changes every clean build. Because this source file feeds into `javac`/`kotlinc` as a compile input, it invalidates the Gradle build cache for every compile and dex task on every build — even when no source code changed. This defeats remote build caching (Develocity/Gradle Enterprise) entirely.

By moving to an asset file, the dynamic values only affect asset merging and APK packaging — cheap I/O-bound tasks that aren't meaningfully cached. Compile and dex tasks stay fully cacheable.

## Changes

**Plugin (Gradle):**
- `NewRelicConfigTask` generates `newrelic_config.json` via `groovy.json.JsonOutput`
- AGP 7.0+ adapters use `variant.sources.assets.addGeneratedSourceDirectory()` instead of `variant.sources.java`
- AGP 4.x adapter wires via `android.sourceSets` asset directories (Why is AGP 4 even supported still?)
- Removed all `addGeneratedSourceDirectory` on Java sources, `registerJavaGeneratingTask`, Kotlin compile task wiring, and `afterEvaluate` blocks tying config to compilation
- Metadata file moved to `intermediates/` to avoid leaking into APK

**Runtime (Agent):**
- `AndroidAgentImpl.loadNewRelicConfig()` reads JSON from assets during init, with distinct handling for missing file (plugin not applied) vs parse errors
- `Agent.setNewRelicConfig()` (package-private) replaces reflection-based loading
- `Agent.getBuildId()` uses double-check pattern to skip synchronized block when the field is already set

**Tests:**
- `NewRelicConfig.installTestConfig()` helper added for agent-core tests (matched package for the visibility to the set function)
- Plugin integration test and 6 functional test specs updated for new output paths

## Notes

**JSON fields:** The generated JSON currently includes all five original fields (`version`, `buildId`, `obfuscated`, `mapProvider`, `metrics`) for parity with the old `NewRelicConfig.java`. However, only `buildId` and `obfuscated` are read at runtime — the map upload task receives its `buildId` and `mapProvider` values directly through Gradle task inputs, not from this file. The unused fields (`version`, `mapProvider`, `metrics`) could be trimmed from the JSON in a follow-up.

**Pre-existing test failures on `develop`:** The `agent-core:test` suite has 21-22 pre-existing failures on `develop` unrelated to this change (Harvester session reset tests, OkHttp3 URL assertion mismatches, HarvestConfiguration serialization, RemoteLogger concurrency). The `plugins:gradle:integrationTests` suite has 1 pre-existing failure (`BuildHelperTest > isUsing7xTransform`). We did not investigate these failures — they reproduce on `develop` without any of our changes applied.


ColtonIdle (commented):
love this change!

agent-core/src/main/java/com/newrelic/agent/android/Agent.java
```diff
@@ -26,6 +25,7 @@ public class Agent {
     private static Object implLock = new Object();
     private static AgentImpl impl = NULL_AGENT_IMPL;
     private static String buildId = null;
```
cmcadams-newrelic: Could we add the `volatile` keyword to `buildId` Which would help with any race conditions. the volatile modifier guarantees that any thread that reads a field will see the most recently written value.

agent-core/src/main/java/com/newrelic/agent/android/Agent.java
```diff
@@ -26,6 +25,7 @@ public class Agent {
...
+    private static boolean obfuscated = false;
```
cmcadams-newrelic: same thing here but for the `obfuscated` keyword

Could we add the `volatile` keyword to `obfuscated` Which would help with any race conditions. the volatile modifier guarantees that any thread that reads a field will see the most recently written value.

cmcadams-newrelic (changes requested):
Nice work. I have two small changes I would like to have addressed before merging.

agent-core/src/main/java/com/newrelic/agent/android/Agent.java
trevjonez: added in [f82166d](https://github.com/newrelic/newrelic-android-agent/pull/496/commits/f82166d5d5e0a398677020443734cbb554fc2ed4)

agent-core/src/main/java/com/newrelic/agent/android/Agent.java
trevjonez: added in [f82166d](https://github.com/newrelic/newrelic-android-agent/pull/496/commits/f82166d5d5e0a398677020443734cbb554fc2ed4)

ndesai-newrelic:
We hit the same root cause independently and landed on a similar fix via a different mechanism:

- `NewRelicConfigTask` writes the real (still random) build ID into a generated Android **string resource** (`values/com_newrelic_android_agent_config.xml`) instead of a JSON asset. `NewRelicConfig.java` stays, but `BUILD_ID` becomes a stable placeholder — so it's cache-safe without removing the compiled class.
- Wired via `variant.sources.res.addGeneratedSourceDirectory(...)` per AGP adapter (4.x/7.0/7.4/9.0).
- Runtime: `AndroidAgentImpl` resolves the value via `Resources.getIdentifier(...)` at init and pushes it to `Agent.setBuildId(...)`; `Agent.getBuildId()` falls back to the existing `NewRelicConfig` reflection path if the resource is missing (covers old-plugin + new-agent-core version skew).

Same fix, different carrier (resource vs. JSON asset) — worth comparing notes since both PRs are touching this at once.

ndesai-newrelic:
@trevjonez we will release it soon

ndesai-newrelic:
@trevjonez could you review and test these changes? https://github.com/newrelic/newrelic-android-agent/pull/590

ColtonIdle:
why closed? @ndesai-newrelic

ndesai-newrelic:
@ColtonIdle we fixed it here https://github.com/newrelic/newrelic-android-agent/pull/590 and will release it soon

PR #501, opened 2026-05-04
[NR-555276]Fix/out of memory issue fix

ndesai-newrelic:


ywang-nr (approved):
I guess this is the POC, will need to double check whether updates from old agents will work if we convert to new way.

ndesai-newrelic:
@ywang-nr this is final implementation , we already going to move from sharedpreferences to file base storage for other data events also. iOS already have file based storage for offline storage.

mbruin-NR (approved):

mbruin-NR (approved):

cdillard-NewRelic (approved):
👍

ndesai-newrelic:
@ywang-nr which updates are you talking about here? we are not migrating older sharedpreferences hex data here and we are also deleting the older NRPayloadStore.

ywang-nr:
> @ywang-nr which updates are you talking about here? we are not migrating older sharedpreferences hex data here and we are also deleting the older NRPayloadStore.

Just want to double check will following process be covered?
1. Application on old agent with SharedPreference way, has som data in it
2. Application updated to new agent with FilePayloadStore
3. While we delete the "NRPayloadStore", will we have data lost? Do we need to have a migration process?

ndesai-newrelic:
@ywang-nr i decided not to migrate and delete whatever data on the sharedpreferences and also mentioned the same on DACI also.

PR #507, opened 2026-05-14
Potential fix for code scanning alert no. 30: Workflow does not contain permissions

ywang-nr:
Potential fix for [https://github.com/newrelic/newrelic-android-agent/security/code-scanning/30](https://github.com/newrelic/newrelic-android-agent/security/code-scanning/30)

In general, the problem is fixed by explicitly specifying a minimal `permissions` block either at the workflow root (applies to all jobs without their own `permissions`) or at the specific job. This prevents the `GITHUB_TOKEN` from inheriting potentially broad default repository/organization permissions and documents the intended privileges.

The best single fix here without changing functionality is to add a *workflow-level* `permissions` block right after the `on:` section. All existing jobs (`setup`, `build-agent`, `test`, `report`) only need to read repository contents (checkout code, read JSON) and write to their own logs/step summaries, which doesn’t require any extra token scopes. The `comment` job already has its own `permissions: pull-requests: write`, and job-level permissions override workflow-level ones, so adding a root `permissions: contents: read` will not interfere with its behavior. Concretely, in `.github/workflows/compatibility-matrix-test.yml`, insert:

```yaml
permissions:
  contents: read
```

between the `on:` block and the `jobs:` key. No additional methods, imports, or definitions are required.


ywang-nr:
Need to change it to develop branch instead of main

ndesai-newrelic (approved):

PR #551, opened 2026-06-03
[NR-548420]feat(session-replay): persist replay payloads when offline

ndesai-newrelic:
Jira ticket: NR-548420

Today, SessionReplaySender silently drops the payload when the network is unreachable and only logs on HTTP failure. Logs, events, handled exceptions, and crashes already persist to disk and retry; session replay didn't.

This adds an OfflineStorage-gated path that:
- Freezes session attributes + URL `timestamp=` at capture time so a retry reproduces the original request (sessionId, replay timestamps, appVersion, instrumentation.*, plus all live session attributes)
- Persists proactively when the network is unreachable (tagged `offline=true` to match Crash/Error/Event/AgentData)
- Persists on retriable failures (408, 429, 5xx, IOException) and drops on hard rejections (400, 403)
- Drains the cache serially in FIFO order on each harvest, halting on first failure to retry next cycle, with a 2-day TTL via agentConfiguration.getPayloadTTL()
- Caps at 50 entries with FIFO eviction via AbstractFileStore

The existing single-entry FileSessionReplayStore is left untouched; offline payloads live in a new nr_offline_session_replay_cache/.


ndesai-newrelic (commented):
Reviewed this change set. No blocking issues found.\n\nThe shutdown-path adjustment in AndroidAgentImpl to clear only pending events (while preserving session/system attributes for the final shutdown harvest) matches the intended behavior, and the added test coverage in AndroidAgentImplTest exercises the regression scenario.

cdillard-NewRelic (approved):
Code looks good but I didn't run it.

PR #556, opened 2026-06-09
NR-572735 added a way to detect KMP and add a supportability metric for it

mbruin-NR:
## Summary
  Adds automatic detection of Kotlin Multiplatform (KMP) usage in Android applications and reports a supportability metric when detected. This enables New Relic to track KMP adoption and provide better support for multiplatform applications https://new-relic.atlassian.net/browse/NR-572735

### Detection Strategy
  The `KmpChecker` uses runtime class detection to identify KMP projects by checking for platform-specific classes:

  - **Kotlin/Native classes** (`kotlin.native.concurrent.*`) - Present when iOS/Native targets are enabled
  - **AtomicFU classes** (`kotlinx.atomicfu.*`) - Used for shared mutable state across platforms
  - **Kotlin/JS classes** (`kotlin.js.*`) - Present when JavaScript/WASM targets are enabled

  ### Key Features
   **No false positives** - Only detects actual KMP projects, not regular Kotlin Android apps
   **Automatic** - Runs during agent initialization
   **Lightweight** - Simple class existence checks with no performance impact

  ### Limitations
  - **False negatives possible**: May not detect Android-only KMP builds (iOS targets disabled)
  - **Most production apps detected**: Real-world KMP apps typically ship with iOS targets enabled

  ## Metric Reported
  When KMP is detected, the agent increments:
  Supportability/Mobile/Android/HybridPlatform/KMP

### DIFF
--- a/agent/src/main/java/com/newrelic/agent/android/util/KmpChecker.java
+++ b/agent/src/main/java/com/newrelic/agent/android/util/KmpChecker.java
@@ -1,0 +1,49 @@
+package com.newrelic.agent.android.util;
+
+public class KmpChecker {
+
+    public static boolean isKmpUsed() {
+        // Check for Kotlin runtime
+        if (!isKotlinPresent()) {
+            return false;
+        }
+
+        // Check for common KMP libraries
+        return hasKmpLibraries();
+    }
+
+    private static boolean isKotlinPresent() {
+        try {
+            Class.forName("kotlin.Unit");
+            return true;
+        } catch (ClassNotFoundException e) {
+            return false;
+        } catch (Throwable e) {
+            return false;
+        }
+    }
+
+    private static boolean hasKmpLibraries() {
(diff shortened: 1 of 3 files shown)


agent/src/main/java/com/newrelic/agent/android/util/KmpChecker.java
```diff
@@ -0,0 +1,49 @@
...
+        } catch (ClassNotFoundException e) {
+            return false;
+        } catch (Throwable e) {
+            return false;
+        }
+    }
+
+    private static boolean hasKmpLibraries() {
+        // Check for coroutines (most KMP apps use this for async operations)
+        boolean hasCoroutines = checkForClass("kotlinx.coroutines.Job");
```
ndesai-newrelic: these libraries are also used for android development also.

ndesai-newrelic (approved):

PR #557, opened 2026-06-09
NR-572828-async-image-compression

ndesai-newrelic:
🔗 Linked to JIRA ticket: [NR-572828](https://newrelic.atlassian.net/browse/NR-572828)
[NR-572828]: https://new-relic.atlassian.net/browse/NR-572828?atlOrigin=eyJpIjoiNWRkNTljNzYxNjVmNDY3MDlhMDU5Y2ZhYzA5YTRkZjUiLCJwIjoiZ2l0aHViLWNvbS1KU1cifQ


ywang-nr (approved):

cdillard-NewRelic:
Potential considerations from Claude. 

- Bitmap thread-safety across the handoff. The bitmap reference is captured on the UI/capture thread and read on compressionExecutor. This is only safe if drawableToBitmap(...) / extractBitmapForPainter(...) return a defensive copy. If they hand back a live/shared bitmap (e.g. BitmapDrawable.getBitmap(), or a Glide/Coil-owned bitmap), the UI thread can recycle or reuse it mid-compression → IllegalStateException/corrupt pixels. The diff doesn't show those methods — please confirm they copy, or copy before dispatch.
- Image resolves one frame late, and may never resolve on a static screen. By design, a cache miss returns null for frame N and only resolves on frame N+1 (cache hit → diff emits background-image). That depends on a subsequent frame being captured. If capture is draw/change-driven and the screen becomes static right after the image appears (image-only screen, then idle), hasChanged may never fire again and the image is never sent. Two questions:
  a. Does the capture cadence guarantee a frame N+1 after compression completes?
  b. Would it be worth having the worker mark the view dirty / request a re-capture on completion to close this gap? The current code only populates the cache.
  - volatile imageData + comment contradict the final design. The field comment says "written off-main," but the worker explicitly does not back-fill imageData (it only writes the cache). If imageData is assigned solely in the constructor on the capture thread, volatile is likely unnecessary and the comment is a stale leftover from a back-filling iteration. Either way the two comments disagree — please reconcile.
- Single-thread executor: unbounded queue + duplicate work. newSingleThreadExecutor uses an unbounded LinkedBlockingQueue. Under a fast capture rate with slow compression, frames N and N+1 both miss (cache not yet populated) and queue two compressions for the same key, holding bitmap references and growing the backlog. Consider an in-flight Set<String> guard (skip dispatch if a key is already queued) and/or a bounded queue to cap memory.
- Executor lifecycle. The static executor is created once and never shut down. Confirm NamedThreadFactory produces a daemon thread (so it can't block process teardown), and consider whether it should be torn down alongside the cache clear on background.

PR #567, opened 2026-06-18
fix(NR-564673): Conditionally call incrementUploadCount 

cmcadams-newrelic:
[NR-564673](https://new-relic.atlassian.net/browse/NR-564673)

## Summary

Fixes a bug where `CrashSender` would increment the upload count and persist the crash to the store **before** attempting the network request — causing the upload counter to advance even when the device was offline or the request failed with no response.

## Problem

The original code called `crash.incrementUploadCount()` and `agentConfiguration.getCrashStore().store(crash)` unconditionally before `super.call()`. If the network request failed (e.g., device offline, `IOException` thrown), the upload count was already incremented, potentially causing crashes to be dropped before they were successfully delivered.

## Solution

Move `incrementUploadCount()` and the store update to execute only after a successful network attempt, guarded by a check that `responseCode > 0`. A `responseCode` of 0 indicates the request never reached the server (e.g., no network connectivity).

## Changes

- `CrashSender.java`: Deferred `incrementUploadCount()` and `crashStore.store()` to only execute when `responseCode > 0` after `super.call()` returns.
- `CrashSenderTest.java`: 
  - Updated `call()` test to remove the pre-test manual `incrementUploadCount()` call that masked the bug.
  - Added `testCallWhenOfflineDoesNotIncrementUploadCount()` to assert the upload count is unchanged when an `IOException` prevents the request from completing.
  - Refactored `getMockedConnection()` helper to remove the unused `CrashSender` parameter.

## Testing

- Existing `call()` test confirms upload count increments on a successful send.
- New `testCallWhenOfflineDoesNotIncrementUploadCount()` test covers the offline/failure scenario.

[NR-564673]: https://new-relic.atlassian.net/browse/NR-564673?atlOrigin=eyJpIjoiNWRkNTljNzYxNjVmNDY3MDlhMDU5Y2ZhYzA5YTRkZjUiLCJwIjoiZ2l0aHViLWNvbS1KU1cifQ

### DIFF
--- a/agent-core/src/main/java/com/newrelic/agent/android/crash/CrashSender.java
+++ b/agent-core/src/main/java/com/newrelic/agent/android/crash/CrashSender.java
@@ -50,12 +50,14 @@ public class CrashSender extends PayloadSender {
     public PayloadSender call() {
         setPayload(crash.toJsonString().getBytes());
 
-        // Increment the upload counter
-        crash.incrementUploadCount();
-        agentConfiguration.getCrashStore().store(crash);
-
         try {
-            return super.call();
+            PayloadSender result = super.call();
+
+            if(responseCode > 0) { // `responseCode` is part of the PayloadSender interface
+                crash.incrementUploadCount();
+                agentConfiguration.getCrashStore().store(crash);
+            }
+            return result;
 
         } catch (Exception e) {
             onFailedUpload("Unable to report crash to New Relic, will try again later. " + e);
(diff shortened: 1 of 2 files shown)


cmcadams-newrelic:
Note: GH action `add-jira-link` will fail until this change is merged https://github.com/newrelic/newrelic-android-agent/pull/567

ndesai-newrelic (approved):

PR #568, opened 2026-06-18
fix: Use pull_request_target for Jira link action on fork PRs

cmcadams-newrelic:
## Summary

Fixes the `add-jira-links` workflow failing with `Resource not accessible by integration` on PRs opened from forks.

## Problem

The `pull_request` event runs the workflow in the context of the **fork**, so GitHub restricts `GITHUB_TOKEN` to read-only — even when `pull-requests: write` is declared in the `permissions` block. This is a GitHub security restriction to prevent malicious forks from gaining write access to the base repo.

## Solution

Switch the trigger from `pull_request` to `pull_request_target`, which runs the workflow in the context of the **base repository**. This gives `GITHUB_TOKEN` the write access needed to update the PR body with the Jira link.

This is safe because `codenamegrant/jira-pr-link-action` only reads the PR title/body and updates it via the GitHub API — it does not check out or execute any code from the fork.

## Changes

- `.github/workflows/jira-link.yml`: `pull_request` → `pull_request_target`

### DIFF
--- a/.github/workflows/jira-link.yml
+++ b/.github/workflows/jira-link.yml
@@ -13,7 +13,7 @@ jobs:
   add-jira-links:
     runs-on: ubuntu-latest
     steps:
-      - uses: codenamegrant/jira-pr-link-action@v1
+      - uses: codenamegrant/jira-pr-link-action@6151a1959fff01a7c73d8bab7081624eb9c4c7ef # v1
         with:
           github-token: ${{ secrets.GITHUB_TOKEN }}
           jira-base-url: 'https://newrelic.atlassian.net'


ndesai-newrelic (approved):

ndesai-newrelic (changes requested):
Nice fix direction — `pull_request_target` is the right trigger to make fork PR updates possible. I do have one blocker before merge:
1) **Please pin `codenamegrant/jira-pr-link-action` to a full commit SHA** (currently `@v1` in `.github/workflows/jira-link.yml`).
Because this now runs under `pull_request_target` with `pull-requests: write`, we are executing a third-party action with elevated repo-context token permissions. Using a mutable tag (`@v1`) leaves us exposed to tag retarget/supply-chain drift. Pinning to an immutable SHA is the standard hardening step here.
Optional hardening follow-up (non-blocking): since there is no checkout, test whether `contents: read` can be dropped (or set to `none`) while keeping the action functional.

ndesai-newrelic (approved):

PR #574, opened 2026-06-24
fix(NR-577137): add TTL expiry and LRU eviction to offline storage

cmcadams-newrelic:
🔗 Linked to JIRA ticket: [NR-577137](https://newrelic.atlassian.net/browse/NR-577137)

## Summary

- Offline storage had no TTL: files persisted indefinitely and could be re-uploaded well past the backend's data retention window
- LRU eviction code existed but was broken — files were never deleted and the running byte total was never updated, making the storage cap effectively unenforced
- Adds a 7-day TTL enforced at read time and fixes the eviction loop to actually free space before rejecting new data

### `OfflineStorage.java`
- Added `DEFAULT_OFFLINE_STORAGE_TTL` (7 days) and  `offlineStorageTTL` getters and setters 
- `getAllOfflineData()`: computes `expiryTime = now - TTL` before iterating files; any file whose `lastModified` predates the expiry is deleted and skipped
- `persistHarvestDataToDisk()`: replaced the early `return false` on capacity overflow with a sorted LRU eviction pass — files are sorted ascending by `lastModified`, deleted oldest-first while subtracting each file's size from `totalData`; returns false only if eviction still cannot bring usage below the cap

### `OfflineStorageTest.java`
- Added `@After tearDown()` to reset static `offlineStorageTTL` and `offlineStorageSize` between tests, preventing ordering dependencies
- New tests: `testDefaultTTL`, `testSetGetTTL`, `testExpiredFilesAreDeletedOnRead`, `testNonExpiredFilesAreReturned`, `testMixedExpiredAndValidFiles`, `testLruEvictionEvictsOldestFile`, `testPersistReturnsFalseWhenEvictionCannotFreeEnoughSpace`
- Fixed pre-existing `testOfflineStorageInstance` failure: `getOfflineFilePath()` returns empty string until a file has actually been persisted

## Test plan

- [ ] Run `./gradlew :agent:testReleaseUnitTest` all `OfflineStorageTest` tests pass
- [ ] Confirm expired files (backdated past 7 days) are deleted and not returned by `getAllOfflineData()`
- [ ] Confirm that when storage is full, oldest files are evicted to make room before rejecting new data
- [ ] Confirm `persistHarvestDataToDisk()` returns `false` only when eviction cannot free sufficient space

[NR-577137]: https://new-relic.atlassian.net/browse/NR-577137?atlOrigin=eyJpIjoiNWRkNTljNzYxNjVmNDY3MDlhMDU5Y2ZhYzA5YTRkZjUiLCJwIjoiZ2l0aHViLWNvbS1KU1cifQ

### DIFF
--- a/agent/src/main/java/com/newrelic/agent/android/util/OfflineStorage.java
+++ b/agent/src/main/java/com/newrelic/agent/android/util/OfflineStorage.java
@@ -10,16 +10,21 @@ import java.io.BufferedWriter;
 import java.io.File;
 import java.io.FileReader;
 import java.io.FileWriter;
+import java.util.Arrays;
+import java.util.Comparator;
 import java.util.HashMap;
 import java.util.Map;
+import java.util.concurrent.TimeUnit;
 
 public class OfflineStorage {
     private static final String OFFLINE_STORAGE = "nr_offline_storage";
     private static final int DEFAULT_MAX_OFFLINE_Storage_SIZE = 100 * 1024 * 1024; //MB
+    static final long DEFAULT_OFFLINE_STORAGE_TTL = TimeUnit.DAYS.toMillis(7);
     private static final AgentLog log = AgentLogManager.getAgentLog();
     private static File offlineStorage;
     private static String offlineFilePath = "";
     private static int offlineStorageSize = 100 * 1024 * 1024; //MB
+    private static long offlineStorageTTL = DEFAULT_OFFLINE_STORAGE_TTL;
 
     public OfflineStorage(Context context) {
         try {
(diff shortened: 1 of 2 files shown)


ndesai-newrelic:
Blocker found in agent/src/main/java/com/newrelic/agent/android/util/OfflineStorage.java (persistHarvestDataToDisk, around lines 43-64):

When totalData exceeds the cap, eviction starts before checking whether the incoming payload can fit by itself. If a new payload is larger than offlineStorageSize, this logic can delete all existing offline files and still return false. That causes data loss of previously queued payloads.

Suggested fix: add an early guard before eviction: if incoming payload size > offlineStorageSize, return false immediately. Optional hardening: only subtract file length from totalData when delete() succeeds.

ndesai-newrelic (approved):

PR #585, opened 2026-07-01
Android Release v7.7.7

ywang-nr:
Ticket: https://new-relic.atlassian.net/browse/NR-577726

What's changed:
1. Multiple bug fixes
- Fixed a performance issue where Session Replay image compression ran on the main thread; it now runs asynchronously, reducing overhead during capture.

- Addressed rendering accuracy issues in Session Replay by emitting incremental view additions in their natural view order.
                                                                                                                                                                                                             
- Fixed an issue where payloads stored offline could be sent later as stale data; offline storage now enforces a time-to-live (TTL) so data queued too long is discarded instead.

- Resolved an issue where session attributes could be lost during the shutdown harvest.

- Fixed a problem where buffered events were dropped instead of being sent in the final harvest as the app shut down, preventing data loss on exit.

- Fixed a timing issue during device rotation where session continuity could be lost during the transition.

- Made HTTP header handling thread-safe, resolving errors that could occur under concurrent access.

- Fixed an issue where app start-up metrics could be miscalculated when a start-up timestamp was unset.

- Resolved an issue where the agent attempted to report crashes without an active network connection, causing unnecessary errors on offline devices.

- Fixed a native (NDK) crash that could occur during shutdown due to an ABI mismatch in `stop()`.

- Fixed ProGuard map upload for builds that produce a DexGuard APK directly.

- Improved Gradle configuration-cache compatibility during map upload.

- Relocated the bundled `commons-io` dependency in the Gradle plugin to prevent conflicts with host applications.


agent-core/src/test/java/com/newrelic/agent/android/background/TestApplicationStateMonitor.java
```diff
@@ -92,6 +92,50 @@ public void shouldNotifyAllListenersWhenInBackground() {
         }
```
ywang-nr, line 92: Update filename: move Test to the end of the filename in next release

agent-core/src/test/java/com/newrelic/agent/android/crash/CrashReporterTest.java
```diff
@@ -15,6 +15,7 @@
 import com.newrelic.agent.android.harvest.Harvest;
```
ywang-nr, line 15: Update the filename: remove 's' from Test's'

gradle.properties
```diff
@@ -8,7 +8,7 @@ android.debug.obsoleteApi=true
 android.useAndroidX=true
 
 # Version of the SDK to be built and deployed
-newrelic.agent.version = 7.7.6
+newrelic.agent.version = 7.7.7-SNAPSHOT
```
ywang-nr: Testing version: -SNAPSHOT, remove before releasing

agent-core/src/test/java/com/newrelic/agent/android/background/TestApplicationStateMonitor.java
ywang-nr: It has duplication in the classes, review in next release

agent-core/src/test/java/com/newrelic/agent/android/crash/CrashReporterTest.java
ywang-nr: Resolved

gradle.properties
ywang-nr: Resolved

ndesai-newrelic (approved):

PR #590, opened 2026-07-15
fix: decouple build ID generation from compile/minify caching (NR-556102)

ndesai-newrelic:
🔗 Linked to JIRA ticket: [NR-556102](https://newrelic.atlassian.net/browse/NR-556102)

## Summary

`NewRelicConfigTask` embeds a fresh random build ID into the app's compiled `NewRelicConfig.java` on every build. Because that generated source feeds `compileReleaseJavaWithJavac`/`compileReleaseKotlin`, and the compiled class then feeds `minifyReleaseWithR8`, a one-line change to an unrelated string constant busts the Gradle build cache for all three tasks on every CI build — even when no source code changed.

This moves the real, still-random build ID out of compiled source and into a generated Android string resource instead, so compile/minify inputs stay stable across builds.

## What & Why

- `NewRelicConfigTask` now writes the build ID into a generated `values/com_newrelic_android_agent_config.xml` string resource (entry `com.newrelic.android.buildId`), named generically so the same file can carry other dynamic values later without another rename.
- `NewRelicConfig.java`'s `BUILD_ID` field becomes a fixed placeholder — the class still exists and compiles the same every build, it just no longer carries the per-build value.
- Each AGP adapter (4.x / 7.0 / 7.4+ / 9.0+) wires the resource directory via `sources.res.addGeneratedSourceDirectory(...)` (legacy `registerGeneratedResFolders` + manual `dependsOn` for 4.x), alongside the existing Java source wiring.
- `AndroidAgentImpl` resolves the value via `Resources.getIdentifier(...)` during `init()` (agent-core has no Android SDK dependency, so this can't happen there) and pushes it to `Agent.setBuildId(...)`.
- `Agent.getBuildId()` returns the pushed value if present, otherwise falls back to its existing `NewRelicConfig` reflection path unchanged — covers a plugin/agent-core version mismatch where an older plugin doesn't generate the resource yet.
- Same root fix as #496, via a different carrier (Android resource vs. JSON asset) — see that PR's discussion for comparison notes.

## Callouts

- **Version skew:** new agent-core + old plugin falls back cleanly (resource absent → reflection path). Old agent-core + new plugin is *not* supported — the old `Agent.java` would reflect the now-placeholder `BUILD_ID`. Plugin and agent-core ship from this repo in lockstep, so this is a documented limitation, not defensive code.
- `CLAUDE.md` gets a short new note under "Build-Time vs Runtime Split" describing this mechanism for future readers (this is also the first commit of `CLAUDE.md` into the repo).

## Test plan

- `./gradlew :plugins:gradle:test` and `:plugins:gradle:integrationTests` — green, including new `NewRelicConfigTaskTest` cases asserting the resource is written and the compiled source keeps the placeholder.
- `./gradlew :plugins:gradle:functionalTests --tests PluginJDK11SmokeSpec --tests PluginJDK17SmokeSpec` — green against a real sample-app build, including the extended "verify NewRelicConfig was injected" assertions.
- `./gradlew :agent-core:test` (new `AgentBuildIdTest`) and `:agent:testReleaseUnitTest` (new `AndroidAgentImplTest` cases) — green, covering both the resource-resolution path and the reflection fallback.
- Manual: two clean builds of `samples/agent-test-app` with `--build-cache` confirm `compileReleaseJavaWithJavac`/`compileReleaseKotlin`/`minifyReleaseWithR8` are `FROM-CACHE` on the second run, while the generated build ID itself still differs between the two builds.

[NR-556102]: https://new-relic.atlassian.net/browse/NR-556102?atlOrigin=eyJpIjoiNWRkNTljNzYxNjVmNDY3MDlhMDU5Y2ZhYzA5YTRkZjUiLCJwIjoiZ2l0aHViLWNvbS1KU1cifQ

### DIFF
--- a/CLAUDE.md
+++ b/CLAUDE.md
@@ -1,0 +1,246 @@
+# New Relic Android Agent
+
+Mobile performance monitoring agent for Android. **Two-phase architecture:** build-time bytecode instrumentation + runtime monitoring and data collection.
+
+## Modules
+
+| Module | Path | Type | Output |
+|---|---|---|---|
+| **agent-core** | `/agent-core` | Java 17 library | `agent-core-<v>.jar` (fat via Shadow) |
+| **instrumentation** | `/instrumentation` | Java 17 library | `instrumentation-<v>.jar` (ASM relocated) |
+| **agent** | `/agent` | Android library | `android-agent-<v>.aar` |
+| **plugins/gradle** | `/plugins/gradle` | Gradle plugin (Groovy + Kotlin) | `agent-gradle-plugin-<v>.jar` |
+
+Dependency chain: `agent` → `agent-core` + `instrumentation`. The Gradle plugin orchestrates build-time rewriting.
+
+### agent-core — runtime agent
+Runtime implementation, public SDK (`NewRelic.java`), data collection, harvest system, analytics, session management.
+
+Key subdirectories:
+```
+agent-core/src/main/java/com/newrelic/agent/android/
+├── crash/                    # Crash detection → .claude/rules/crash-flow.md
+├── tracing/                  # Activity/interaction tracing → .claude/rules/tracing.md
+├── harvest/                  # Data harvesting and transmission
+├── measurement/              # Performance metrics
+├── instrumentation/          # HTTP/network runtime hooks
+├── analytics/                # Events, breadcrumbs
+├── sessionReplay/            # Session replay
+└── distributedtracing/       # Trace context propagation
+```
+
+Key classes: `Agent` (singleton facade), `AndroidAgentImpl` (main impl), `AgentConfiguration`, `Harvest`, `TraceMachine`.
+
+### agent — Android wrapper
+Combines agent-core + instrumentation into an AAR. Handles Android Context integration, lifecycle callbacks, NDK bridge.
+
+```
+agent/src/main/java/com/newrelic/agent/android/
+├── instrumentation/          # Runtime hooks (OkHttp, etc.)
+├── ndk/                       # Native crash bridge
+├── rum/                       # Real User Monitoring (Activity lifecycle)
+├── stores/                    # SharedPreferences persistence
+├── sessionReplay/             # Session replay rendering
+└── webView/                   # WebView instrumentation
+```
+
+Fat JAR lands at `build/intermediates/aar_main_jar/release/classes.jar`. Shadow plugin uses multi-phase JAR build for Gson compatibility (unshaded + shaded → merged).
+
+### instrumentation — bytecode rewriter
+See **`.claude/rules/instrumentation.md`**.
+
+### plugins/gradle — build-time plugin
+See **`.claude/rules/plugin-architecture.md`**.
+
+## Path-Gated Rule Files
+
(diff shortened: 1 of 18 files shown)


trevjonez:
I think this will work better than things do today, but I still don't love it. Using strings still implicates more downstream tasks than an asset file would.

Plus leaving the unused metrics field in the generated java class means cache entries from different machines are not likely to be usable. IE: local build won't be able to use an entry produced by a CI job.

Metrics isn't even used at runtime as far as I can see, can we just drop that from the generated java?

ndesai-newrelic:
@trevjonez i can remove unused metrics field for now.

trevjonez:
If you can, will you check to see what the incremental impact is with this being a string resource?

I don't remember off the top of my head how the resource shrinking sits in the task graph from AGP. I believe AGP 9 enables resource shrinking by default which I think is a part of the R8 task. 

Obviously the vast majority of builds people are doing are just going to be D8 rather than R8, but worth considering while you work through all this for us.

ndesai-newrelic:
@trevjonez good catch, and it turned out to be a real problem, not just a theoretical one. Checked it directly: with `shrinkResources true` (which I understand is on by default under AGP 9), AAPT2's resource shrinker only does static reachability analysis — it can't see through `Resources.getIdentifier()`, which is the only way the agent reads these strings. So `com_newrelic_android_buildId`/`com_newrelic_android_metrics` were getting silently stripped from the release APK. That's worse than the pre-PR state: since `NewRelicConfig.BUILD_ID` is now a permanent placeholder, a stripped resource meant `Agent.getBuildId()` returned an empty string with no working fallback at all — breaking crash-to-mapping correlation for exactly the kind of size-optimized release build that cares about this PR the most.

Fixed it by shipping a static `res/raw/newrelic_keep.xml` inside the agent AAR itself with a `tools:keep` rule for both strings ([`12bdcd9e`](https://github.com/newrelic/newrelic-android-agent/commit/12bdcd9e)). A couple of reasons this is better than generating it from the plugin task:
- The resource *names* are fixed constants (only the *values* differ per build), so a static file protects every consumer regardless of which plugin version generated the value resource — no version-skew risk between agent-core and the Gradle plugin.
- Named it distinctly rather than the conventional `keep.xml`, since apps commonly run this agent alongside Crashlytics (which ships exactly `res/raw/keep.xml`) — reusing that name would risk an AAPT2 duplicate-resource conflict between the two libraries.

One more wrinkle while verifying this: AAPT2's `tools:keep`/`tools:discard` matching works against the Java-safe symbol table, which canonicalizes periods to underscores — so `tools:keep="@string/com.newrelic.android.buildId"` silently didn't match anything even though the runtime `getIdentifier()` call with the dotted name works fine. Standardized on underscores everywhere the name is declared or referenced (XML declaration, runtime lookup, keep rule) rather than keeping two spellings of the same resource around ([`6cae6fac`](https://github.com/newrelic/newrelic-android-agent/commit/6cae6fac)).

Also dropped the unused `METRICS` field from the generated `NewRelicConfig.java` per your earlier comment — the value now only lives in the generated resource, same treatment as build ID.

ywang-nr (approved):

PR #603, opened 2026-08-04
fix(aei): Persist backgrounded state at exit time for accurate AEI background attribute

cmcadams-newrelic:
## Summary

Fixes a bug where AEI (ApplicationExitInfo) derived errors — native crashes, ANRs, OS-initiated kills — were systematically missing the `background=true` attribute even when the original exit occurred while the app was backgrounded.

**Root cause:** `Error.getErrorSessionAttributes()` was calling `ApplicationStateMonitor.isAppInBackground()` at the moment the `Error` object was constructed. That construction happens inside `harvestApplicationExitInfo()`, which runs on the *next* app launch — at which point the app has just foregrounded, so `isAppInBackground()` is almost always `false` regardless of the historical exit state. `AEISessionMeta` had no way to recover the correct value because it only stored `sessionId` and `realAgentId`.

**Fix:** Persist the app's actual foreground/background state at exit time into `AEISessionMeta.backgrounded`, and use that persisted value in `Error` when stamping the attribute — instead of live `ApplicationStateMonitor` state. The `BackgroundReporting` feature flag still gates the behavior.

## Changes

- **`AEISessionMeta`** — adds a `backgrounded` boolean field; all construction sites updated
- **`AEISessionMapper`** — adds `getAppBackgrounded(int pid)`; renames `load()` → `restore()` for clarity; fixes the raw-type cast in the Gson deserialization path; changes return type of `restore()` to `void`
- **`Error`** — replaces the live `ApplicationStateMonitor.isAppInBackground()` check with `sessionMeta.backgrounded` when `BackgroundReporting` is enabled; also increments `BACKGROUND_CRASH_COUNT` metric on background exits
- **`ApplicationExitMonitor`** — calls `restore()` instead of `load()`; moves `sessionMeta` declaration outside the loop so flush uses the correct value
- **Tests** — adds `getAppBackgrounded` test; renames `load` test to `restore`; adds 6 new `ErrorTests` covering: backgrounded=true adds attribute, backgrounded=false omits attribute, feature-flag disabled suppresses attribute, metric increment, persisted state overrides live state, null meta leaves attribute unchanged

## Test plan

- [ ] Run `./gradlew :agent-core:test :agent:test` — all new and existing AEI/Error tests pass
- [ ] Verify `background=true` appears on AEI error events for exits that occurred while backgrounded
- [ ] Verify `background` attribute is absent when `BackgroundReporting` feature flag is disabled
- [ ] Verify no regression on live (non-AEI) crash/handled-exception background attribute stamping

Fixes [NR-596838](https://new-relic.atlassian.net/browse/NR-596838)

[NR-596838]: https://new-relic.atlassian.net/browse/NR-596838?atlOrigin=eyJpIjoiNWRkNTljNzYxNjVmNDY3MDlhMDU5Y2ZhYzA5YTRkZjUiLCJwIjoiZ2l0aHViLWNvbS1KU1cifQ

### DIFF
--- a/agent-core/src/main/java/com/newrelic/agent/android/background/ApplicationStateMonitor.java
+++ b/agent-core/src/main/java/com/newrelic/agent/android/background/ApplicationStateMonitor.java
@@ -108,6 +108,7 @@ public class ApplicationStateMonitor {
     }
 
     private void notifyApplicationInForeground() {
+        foregrounded.set(true);
         final ArrayList<ApplicationStateListener> listeners;
         synchronized (applicationStateListeners) {
             listeners = new ArrayList<ApplicationStateListener>(applicationStateListeners);
@@ -116,6 +117,7 @@ public class ApplicationStateMonitor {
         for (ApplicationStateListener listener : listeners) {
             listener.applicationForegrounded(e);
         }
+
     }
 
     public ExecutorService getExecutor() {
@@ -126,7 +128,7 @@ public class ApplicationStateMonitor {
         return foregrounded.get();
     }
 
-    public static boolean isAppInBackground() {
+    public static boolean isBackgrounded() {
         return !getInstance().getForegrounded();
     }
 
--- a/agent-core/src/main/java/com/newrelic/agent/android/aei/AEISessionMapper.java
+++ b/agent-core/src/main/java/com/newrelic/agent/android/aei/AEISessionMapper.java
@@ -55,10 +55,6 @@ public class AEISessionMapper {
         return model == null ? "" : model.sessionId;
     }
 
-    public boolean getAppBackgrounded(int pid) {
-        AEISessionMeta model = get(pid);
-        return model != null && model.backgrounded;
-    }
 
     public int getRealAgentID(int pid) {
         AEISessionMeta model = get(pid);
--- a/agent-core/src/main/java/com/newrelic/agent/android/aei/AEISessionMapper.java
+++ b/agent-core/src/main/java/com/newrelic/agent/android/aei/AEISessionMapper.java
@@ -67,7 +67,7 @@ public class AEISessionMapper {
     }
 
     @SuppressWarnings("unchecked")
-    public AEISessionMapper load() {
+    public void load() {
         if (mapStore.exists() && mapStore.canRead()) {
             try {
                 String storeData = Streams.slurpString(mapStore, StandardCharsets.UTF_8.toString());
@@ -82,8 +82,6 @@ public class AEISessionMapper {
         } else {
             AgentLogManager.getAgentLog().debug("Cannot read session ID mapper: file does not exist or is unreadable");
         }
-
-        return this;
     }
 
     public boolean flush() {
@@ -133,11 +131,14 @@ public class AEISessionMapper {
     public static class AEISessionMeta {
         final String sessionId;
         public final int realAgentId;
+        public boolean backgrounded;
 
-        public AEISessionMeta(String sessionId, int realAgentId) {
+        public AEISessionMeta(String sessionId, int realAgentId, boolean backgrounded) {
             this.sessionId = sessionId == null ? "" : sessionId;
             this.realAgentId = realAgentId;
+            this.backgrounded = backgrounded;
         }
+
 
         public boolean isValid() {
             return !(sessionId.isEmpty() || realAgentId == 0);
--- a/agent-core/src/main/java/com/newrelic/agent/android/aei/AEISessionMapper.java
+++ b/agent-core/src/main/java/com/newrelic/agent/android/aei/AEISessionMapper.java
@@ -32,7 +32,7 @@ public class AEISessionMapper {
         this.mapStore = mapStore;
         this.mapper = new HashMap<>();
         if (mapStore.exists()) {
-            load();
+            restore();
         }
     }
 
@@ -55,6 +55,11 @@ public class AEISessionMapper {
         return model == null ? "" : model.sessionId;
     }
 
+    public boolean getAppBackgrounded(int pid) {
+        AEISessionMeta model = get(pid);
+        return model != null && model.backgrounded;
+    }
+
     public int getRealAgentID(int pid) {
         AEISessionMeta model = get(pid);
         return model == null ? 0 : model.realAgentId;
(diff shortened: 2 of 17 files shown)


gradle.properties
```diff
@@ -8,7 +8,7 @@ android.debug.obsoleteApi=true
 android.useAndroidX=true
 
 # Version of the SDK to be built and deployed
-newrelic.agent.version = 7.8.0
+newrelic.agent.version = 7.9.0-SNAPSHOT
```
ndesai-newrelic: it should be 7.8.1

ndesai-newrelic (commented):
## Summary
The fix addresses a real bug (live `ApplicationStateMonitor` state is wrong by the time `harvestApplicationExitInfo()` runs), but there's a design gap that likely reproduces the same class of bug one level down, plus a metric double-counting issue.

## 🔴 Correctness — the persisted `backgrounded` value is still captured at the wrong time

`ApplicationExitMonitor.harvestApplicationExitInfo()` is only invoked from `AndroidAgentImpl.onHarvestConnected()` — a **once-per-launch** callback, not a periodic one:

```java
new ApplicationExitMonitor(context).harvestApplicationExitInfo();
```

At the end of that method, the *current* session's `AEISessionMeta` (the one that will be read back on the **next** launch, describing however this session eventually dies) is built like this:

```java
boolean isBackgrounded = ApplicationStateMonitor.isAppInBackground();
AEISessionMapper.AEISessionMeta model = new AEISessionMapper.AEISessionMeta(..., isBackgrounded);
sessionMapper.put(getCurrentProcessId(), model);
sessionMapper.flush();
```

This is written **once, right after launch** (near foreground almost by definition), and nothing else ever updates it — there's no listener hooking foreground/background transitions to re-`put`/`flush` the meta for the running process. So if the app backgrounds later in the session and is then killed/ANRs/crashes while backgrounded, the *next* launch will still read `backgrounded=false` for that exit — the exact symptom this PR set out to fix, just deferred by one generation. To actually work, the current-session's `AEISessionMeta.backgrounded` needs to be refreshed whenever `ApplicationStateMonitor`'s fg/bg state changes (or at minimum re-flushed periodically), not stamped once at harvest-connected time.

## 🟠 Correctness — `BACKGROUND_CRASH_COUNT` can be double-counted or go out of sync with the attribute

`Error.getErrorSessionAttributes()` (unchanged, still called unconditionally from the constructor) independently stamps the background attribute/metric from **live** state:

```java
if (FeatureFlag.featureEnabled(FeatureFlag.BackgroundReporting)) {
    if (ApplicationStateMonitor.isAppInBackground()) {
        attrs.add(new AnalyticsAttribute(AnalyticsAttribute.BACKGROUND_ATTRIBUTE_NAME, true));
        StatsEngine.notice().inc(MetricNames.BACKGROUND_CRASH_COUNT);   // <-- (A)
    }
}
```

Then the new sessionMeta-aware block in the constructor removes/re-derives the *attribute* from persisted state, but doesn't touch the metric already incremented at (A):

```java
if (FeatureFlag.featureEnabled(FeatureFlag.BackgroundReporting) && this.sessionAttributes != null) {
    this.sessionAttributes.removeIf(attr -> AnalyticsAttribute.BACKGROUND_ATTRIBUTE_NAME.equals(attr.getName()));
    if (sessionMeta.backgrounded) {
        this.sessionAttributes.add(new AnalyticsAttribute(AnalyticsAttribute.BACKGROUND_ATTRIBUTE_NAME, true));
        StatsEngine.notice().inc(MetricNames.BACKGROUND_CRASH_COUNT);   // <-- (B)
    }
}
```

- If live state is background but `sessionMeta.backgrounded == false`: metric incremented at (A) but attribute ends up absent — metric says "background crash," attribute says otherwise.
- If both live and persisted are `true`: metric increments **twice** at (A) and (B) for a single error event.

Only the test that covers the metric (`backgroundedTrueIncrementsMetric`) forces live state to foreground, so it never exercises either divergence. Given the persisted value is meant to be authoritative here, the live-state stamping (attribute *and* metric) should probably be skipped entirely when `sessionMeta != null`, not just have its attribute overwritten afterward.

## 🟡 Minor / style

- **Dead hoist**: `AEISessionMapper.AEISessionMeta sessionMeta = null;` was moved above the loop in `ApplicationExitMonitor.java`, but nothing after the loop reads it — the final `model` uses live `ApplicationStateMonitor.isAppInBackground()`, not `sessionMeta`. The PR description says this hoist is "so flush uses the correct value," but that's not what the code does — looks like leftover from an earlier iteration of the fix.
- **Unused new API**: `AEISessionMapper.getAppBackgrounded(int pid)` is added but not called anywhere in production code (only from its own test) — every real call site reads `sessionMeta.backgrounded` directly off the `AEISessionMeta` it already has.
- **Public API rename**: `AEISessionMapper.load()` (returned `this`, fluent) → `restore()` (returns `void`). It's an internal package and the one caller was updated, but it's a public method rename on a public class in `agent-core` — worth double-checking no hybrid-agent code (RN/Flutter/MAUI) reaches into agent-core internals via reflection here.
- **Scope creep**: `PayloadControllerTest.java` and `SessionReplayReporterTest.java` changes (final fields, `assertEquals` arg-order fixes, `@SuppressWarnings`) are unrelated to the AEI background fix. Harmless, but might be cleaner in a separate commit/PR.

## Performance / Lifecycle
- No new O(n²) work; `restore()`'s `forEach`/`putIfAbsent` is the same O(n) as before, just with the raw-type cast removed (a genuine fix for an unchecked-cast smell, not just style).
- No `Context`/`Activity` references are retained by `AEISessionMeta` or `Error`; nothing here risks a leak.

**Bottom line**: intent is right, but I'd hold this for revision — the core fix (deferred write of `backgrounded`) doesn't cover the common "backgrounded later in a long-running session" case, and the metric can go out of sync with the attribute it's meant to describe.

agent-core/src/main/java/com/newrelic/agent/android/background/ApplicationStateMonitor.java
```diff
@@ -126,7 +128,7 @@ public boolean getForegrounded() {
         return foregrounded.get();
     }
 
-    public static boolean isAppInBackground() {
+    public static boolean isBackgrounded() {
```
ndesai-newrelic, line 131: why are you changing function name?

cmcadams-newrelic: looking through the file I thought `isBackgrounded()` would be more consistent with the other property/method names in the class. ie( `foregrounded.set(true)`)

PR #610, opened 2026-08-18
fix(lifecycle): Updates foregrounded flag before listener and harvest

cmcadams-newrelic:
**Updates foregrounded flag before listener and async harvest**

- `ApplicationStateMonitor.uiHidden()`: Move foregrounded.set(false) before `notifyApplicationInBackground()`
- `AndroidAgentImpl.applicationBackgrounded()`: Replace blocking **stop() harvestNow(true,true)** with **stop(false) + async Harvest.harvestNow**
- `AndroidAgentImpl.applicationForegrounded()`: Follows same pattern as `applicationBackgrounded`, and  guarded by !`isAppInBackground()` and `!isShutdown`
- `TestApplicationStateMonitor`: Test rapidCyclingSettlesOnCorrectFinalState


ndesai-newrelic:
## Review

Nice fix on `ApplicationStateMonitor.uiHidden()` — flipping `foregrounded` before notifying listeners is correct, and `foregroundedFlagIsFlippedBeforeBackgroundListenerFires` pins that ordering well.

However, I think there's a critical regression in `AndroidAgentImpl`:

### 🔴 `stop(false)` already shuts Harvest down before the async `harvestNow(true, callback)` can run

In `AndroidAgentImpl.stop(boolean finalSendData)`:

```java
if (finalSendData) {
    ...
    Harvest.harvestNow(true, true);
    ...
}
...
Harvest.shutdown();   // <-- runs UNCONDITIONALLY, regardless of finalSendData
```

`Harvest.shutdown()` is outside the `if (finalSendData)` block, so it runs even when we call `stop(false)`. It nulls out `Harvest`'s internal `harvester`/`harvestTimer` via `shutdownHarvester()`.

Both `applicationForegrounded()` and `applicationBackgrounded()` now do:

```java
stop(false);
Harvest.harvestNow(true, () -> { ...start()/restart logic... });
```

By the time `Harvest.harvestNow(...)` runs, `Harvest.isInitialized()` is already `false` (because `stop(false)` just called `Harvest.shutdown()`), and `harvestNow` opens with `if (isInitialized())`. So the whole call — including `harvestTimer.tickNow(onComplete)` — is a no-op, and the `onComplete` callback (which calls `start()`, restores `BACKGROUND_ATTRIBUTE_NAME`, restarts log reporting) is silently discarded and never runs.

Concrete impact:
- With `FeatureFlag.BackgroundReporting` enabled, the first time the app backgrounds, the whole agent (Harvest, Sampler, TraceMachine, AnalyticsController, Measurements, PayloadController, SessionReplay, LogReporting) is torn down and **never restarts** — the feature this PR is meant to preserve is fully broken.
- Independent of that flag: the old blocking flush that used to happen via `stop(true)`'s `Harvest.harvestNow(true, true)` no longer happens on background, and its async replacement is a no-op — so any buffered HTTP transactions/events/metrics not yet harvested are now silently dropped on every backgrounding.

Suggested fix direction: don't let `stop(false)` fully tear down `Harvest` before the intended async harvest+restart has a chance to run — e.g. have the async path trigger the harvest tick and only shut down (or reinitialize via `start()`) once that tick actually completes, rather than shutting down synchronously first and then trying to schedule work against the now-dead instance.

### Minor: duplicated pattern

`applicationForegrounded()` and `applicationBackgrounded()` both inline the same `stop(false); Harvest.harvestNow(true, () -> {...});` shape. Once the above is fixed, worth extracting a small shared helper so a future change to one side doesn't silently diverge from the other.

### Not applicable
This PR only touches Java (`ApplicationStateMonitor.java`/test, `AndroidAgentImpl.java`) — no Kotlin code, so idiomatic-Kotlin review doesn't apply here. Good to see it reuses the existing `Harvest.harvestNow(boolean, Runnable)` / `HarvestTimer.tickNow(Runnable)` overloads (already on `develop` from NR-552108) rather than adding new plumbing.

PR #615, opened 2026-08-28
fix: derive content length via bounded peekBody for headerless OkHttp3 responses

ndesai-newrelic:
## Jira

- https://new-relic.atlassian.net/browse/NR-611602
- https://new-relic.atlassian.net/browse/NR-610732 (customer support case)

## What & Why

HTTP/2 responses (e.g. Ktor's OkHttp engine) omit `Content-Length` entirely, since HTTP/2 uses binary framing instead of that header. After the 7.7.1 change that removed the `peekBody()` fallback (to stop ~100s hangs on streaming/SSE responses), any headerless response fell back to `bytesReceived = 0` — causing a customer to see ~72% of their `MobileRequest` events report 0 bytes after upgrading to 7.7.7.

This restores a content-length fallback for `OkHttp3TransactionStateUtil.exhaustiveContentLength()`, but bounded and guarded so it can't reintroduce the original hang:

- Falls back to `response.peekBody(512 * 1024L)` only when content length is still unknown after checking the body and both the response's and network response's `Content-Length` headers.
- The 512KB cap bounds memory use for large bodies (e.g. images) — result is capped rather than exposing the true size beyond that.
- Skips the peek entirely for `text/event-stream` (SSE) responses and WebSocket upgrade responses (`Sec-WebSocket-Accept` present), since peeking those can block indefinitely on a live/slow-trickling stream — this preserves the hang-avoidance the original 7.7.1 change was for.

## Callouts

- Only the OkHttp3 path is touched. OkHttp2 already has an equivalent `PrebufferedResponseBody` wired into its instrumentation path and isn't affected by this regression.
- Several pre-existing tests asserted `-1` (unknown length) for headerless, non-streaming responses — that was the bug itself, so those assertions were updated to the correct resolved length as part of this fix.

## Test plan

- `agent-core/src/test/java/com/newrelic/agent/android/instrumentation/okhttp3/OkHttp3TransactionStateUtilTest.java`:
  - New: bounded peek resolves real length for a headerless (HTTP/2-style) response.
  - New: peek is skipped (length stays `-1`) for `text/event-stream` responses.
  - New: peek is skipped (length stays `-1`) for WebSocket upgrade responses.
  - New: peek is capped at exactly 512KB one byte over the boundary; resolves exact size at the boundary.
  - Updated existing tests that previously locked in the `-1` bug for non-streaming headerless responses.


agent-core/src/main/java/com/newrelic/agent/android/instrumentation/okhttp3/OkHttp3TransactionStateUtil.java
```diff
@@ -127,13 +133,39 @@ private static long exhaustiveContentLength(Response response) {
...
         }
 
-        // NOTE: Do NOT use peekBody() to determine content length for responses with unknown length.
-        // peekBody() can block indefinitely for streaming responses (SSE, chunked transfers, etc.)
-        // causing ~100 second delays. It's acceptable for content length to remain unknown (-1).
+        // If still unknown (e.g. HTTP/2 responses, which omit Content-Length entirely), fall
+        // back to a bounded peekBody() read. Skip this for streaming (SSE) and WebSocket
+        // upgrade responses: peekBody() blocks until MAX_BODY_PEEK bytes arrive or the stream
+        // ends, which can take ~100 seconds on a live, slow-trickling stream.
+        if (contentLength < 0L && response != null && !isStreamingOrWebSocketResponse(response)) {
```
cdillard-NewRelic, line 140: does iOS need any changes for parity here?

cdillard-NewRelic (approved):
👍

agent-core/src/main/java/com/newrelic/agent/android/instrumentation/okhttp3/OkHttp3TransactionStateUtil.java
ndesai-newrelic: i dont think so, I am fixing bug here.

PR #621, opened 2026-09-02
Nr 382848 remove nr headers

cmcadams-newrelic:
Removes the `newrelic` header from Distributed Tracing calls. 
Includes tests to validate the `newrelic` header is not present.


samples/agent-test-app/src/main/java/com/newrelic/agent/android/testapp/MainActivity.java
```diff
@@ -69,7 +69,7 @@ protected void onCreate(Bundle savedInstanceState) {
         NewRelic.enableFeature(FeatureFlag.BackgroundReporting);
 
 
-        NewRelic.withApplicationToken("<APP-ID>")    // ## Provide application ID to test map uploads
+        NewRelic.withApplicationToken("AA98d0bba8eecd0aa92f34e3dc9f4eb26c2fec2603-NRMA")    // ## Provide application ID to test map uploads
```
ndesai-newrelic: cant pass application toke, it could be security incident

agent-core/src/test/java/com/newrelic/agent/android/distributedtracing/TraceContextTest.java
```diff
@@ -129,7 +126,7 @@ public void testAsTraceAttributes() {
     public void testGetPayloadTraceHeader() {
         TraceHeader traceHeader = traceContext.getTracePayload();
         Assert.assertEquals(TracePayload.TRACE_PAYLOAD_HEADER, traceHeader.getHeaderName());
-        Assert.assertNotNull(traceHeader.getHeaderValue());
+        Assert.assertFalse(traceContext.getHeaders().stream().anyMatch(h -> TracePayload.TRACE_PAYLOAD_HEADER.equals(h.getHeaderName())));
```
ndesai-newrelic, line 129: **Test coverage regression**: this replaces the old `Assert.assertNotNull(traceHeader.getHeaderValue())` with a verbatim copy of the assertion already added in `testGetHeaders()` above. As written, this test no longer exercises `TracePayload.getHeaderValue()`/`asBase64Json()` at all, and is now redundant with `testGetHeaders()`. Since `getTracePayload()` still returns a live payload object (used by `asTraceAttributes()`), its serialization path is left completely untested. Suggest keeping (or restoring) a direct assertion on `traceHeader.getHeaderValue()` here in addition to the "not in getHeaders()" check.

agent-core/src/test/java/com/newrelic/agent/android/distributedtracing/TraceContextTest.java
```diff
@@ -49,11 +49,8 @@ public void testCreateTraceContext() {
 
     @Test
     public void testGetHeaders() {
-        Assert.assertNotNull(traceContext.tracePayload);
-        Assert.assertFalse(traceContext.getHeaders().isEmpty());
-
-        traceContext.legacyHeadersEnabled = false;
         Assert.assertFalse(traceContext.getHeaders().isEmpty());
+        Assert.assertFalse(traceContext.getHeaders().stream().anyMatch(h -> TracePayload.TRACE_PAYLOAD_HEADER.equals(h.getHeaderName())));
```
ndesai-newrelic, line 53: Nit — style consistency: this introduces `.stream().anyMatch(...)` for a Set membership check, but `testHeadersContainsSpanId()` a few lines below (line 66) does the equivalent check with a plain `for` loop, and the rest of agent-core avoids streams/lambdas in favor of that older iteration style. Consider reusing the same for-loop idiom already used in this file instead of introducing a second way to do the same check (same comment applies to the duplicate at line 129).

agent-core/src/main/java/com/newrelic/agent/android/distributedtracing/TraceContext.java
```diff
@@ -151,13 +145,11 @@ public static void reportSupportabilityExceptionMetric(Exception e) {
     }
 
     public Map<String, Object> asTraceAttributes() {
-        final Map<String, Object> attributeMap = new HashMap<String, Object>() {{
+        return new HashMap<String, Object>() {{
```
ndesai-newrelic, line 148: Nit: `getHeaders()` right above this was just simplified away from the anonymous double-brace-init `HashSet` subclass to a plain `new HashSet<>()` — nice cleanup. `asTraceAttributes()` still uses the same double-brace-init anti-pattern for the `HashMap`. Since this file is already open for this exact kind of cleanup, might as well flatten this one too (plain `HashMap` + `put` calls, or `Map.of(...)`).

agent/src/test/java/com/newrelic/agent/android/aei/ApplicationExitMonitorTest.java
```diff
@@ -262,13 +262,9 @@ public void aeiHarvestShouldNotTriggerEventHarvest() {
 
         loadSessionMapper();
         applicationExitMonitor.harvestApplicationExitInfo();
```
ndesai-newrelic, line 264: This hunk (deleting dead commented-out assertions) is unrelated to the DT header removal that's the point of this PR. Not objecting to the cleanup itself, but consider splitting unrelated cleanups like this (and the `settings.gradle` trailing-line change) into their own commit/PR so the history stays easy to bisect.

settings.gradle
```diff
@@ -43,4 +43,3 @@ include ':agent'
 include ':agent-core'
 include ':instrumentation'
 include ':plugins:gradle'
```
ndesai-newrelic, line 45: This just drops a trailing blank line — looks like fallout from the "chore: revert gradle version" commit rather than something intentional to this PR. Same scope-creep note as the `ApplicationExitMonitorTest` change: fine to keep, but unrelated to the stated purpose of this PR.

ndesai-newrelic (commented):
Reviewed the header-removal logic in `TraceContext`/`W3CTraceContext` — the core change (excluding `tracePayload` from `getHeaders()`) is correct and low-risk: no perf regression (the Gson/base64 serialization in `TracePayload.getHeaderValue()` simply never runs now since it's never iterated), and no lifecycle-safety concerns since agent-core has no Android `Context`/`Activity` coupling.

Two things I'd want addressed before merge (left as inline comments), plus a few nits:

- **Stale Javadoc outside this diff**: `TracePayload.java`'s class doc still says the `newrelic` header "is still sent to maintain backward compatibility," and `DistributedTracing.java`'s header comment still describes three headers being sent (including `newrelic`). Both are now incorrect and directly contradict this PR's purpose. Worth a quick follow-up doc fix, even though neither file is touched by this diff so I can't anchor a line comment on them here.
- If this is meant to be a staged **deprecation** (per the PR description's wording) rather than a hard removal, consider marking `TracePayload`/its header constant `@Deprecated` so consumers get a compile-time signal, rather than silently dropping it from `getHeaders()`.

Inline notes below on the rest.

ndesai-newrelic:
Filed the javadoc follow-up from the review as a separate PR: #628

agent/src/test/java/com/newrelic/agent/android/aei/ApplicationExitMonitorTest.java
cmcadams-newrelic: that's fair I'll split this change out to a seperate PR

settings.gradle
cmcadams-newrelic: removing additional line.

ndesai-newrelic (approved):

PR #623, opened 2026-09-04
fix(session-replay): confine Curtains access to the main thread

ndesai-newrelic:
## Jira
[NR-608999](https://new-relic.atlassian.net/browse/NR-608999)

## What & Why
`SessionReplay.startRecording()`/`stopRecording()` touched `curtains.Curtains` from two different threads at once:
- `startRecording()` posted `Curtains.getRootViews()` to the main thread via `uiThreadHandler.post{}`, but called `Curtains.getOnRootViewsChangedListeners().add(...)` synchronously on the caller's thread.
- `stopRecording()` had the same shape: `viewDrawInterceptor.stopIntercept()` was posted to the main thread, but `Curtains.getOnRootViewsChangedListeners().clear()` ran on the caller's thread immediately.

Both `getRootViews()` and `getOnRootViewsChangedListeners()` resolve the same `rootViewsSpy` delegate inside Curtains, which uses `LazyThreadSafetyMode.NONE` (compiles to `kotlin.UnsafeLazyImpl`, whose internal fields are not volatile). When `startRecording()` runs on an agent background thread (`NR_Harvester-1`, reached via `Harvester.execute()` → `AndroidAgentImpl` → `SessionReplay.startRecording()`), this two-thread access races on that lazy and crashes the host app with an NPE at `kotlin.UnsafeLazyImpl.getValue`. Curtains is a documented main-thread-only, non-thread-safe API.

## Fix
Fold every Curtains call into the existing `uiThreadHandler.post{}` block in both methods, so all access is confined to the main thread:
- `startRecording()`: listener registration moved inside the post block, ahead of the `decorViews.length == 0` early return (so cold start, with no root view yet, still registers the listener).
- `stopRecording()`: now unconditionally posts to the main thread; the `viewDrawInterceptor != null` guard moved inside the runnable so `Curtains.getOnRootViewsChangedListeners().clear()` always runs on the main thread.

## Callouts
- No callers depend on the previous synchronous listener add/clear behavior (checked all internal call sites in `SessionReplay.java` and `AndroidAgentImpl.java`).
- `startRecording()`/`stopRecording()` had zero prior unit test coverage.

## Test plan
- [Diff](https://github.com/newrelic/newrelic-android-agent/pull/new/NR-608999-curtains-threading-race)
- Added `SessionReplayCurtainsThreadingTest`: calls `startRecording()`/`stopRecording()` from a real background thread against a paused Robolectric main `Looper`, and asserts Curtains' listener list is unchanged immediately after the call returns, only mutating once the paused `Looper` is idled — i.e. all Curtains access is confined to the main thread. Against the pre-fix code this assertion fails (listener mutates immediately on the background thread).

[NR-608999]: https://new-relic.atlassian.net/browse/NR-608999?atlOrigin=eyJpIjoiNWRkNTljNzYxNjVmNDY3MDlhMDU5Y2ZhYzA5YTRkZjUiLCJwIjoiZ2l0aHViLWNvbS1KU1cifQ

### DIFF
--- a/agent/src/main/java/com/newrelic/agent/android/sessionReplay/SessionReplay.java
+++ b/agent/src/main/java/com/newrelic/agent/android/sessionReplay/SessionReplay.java
@@ -314,7 +314,20 @@ public class SessionReplay implements OnFrameTakenListener, HarvestLifecycleAwar
             persistSrState(true, true);
         }
 
+        // curtains.Curtains is a main-thread-only, non-thread-safe API (NR-608999): every
+        // access must happen inside this single post() so the calling thread never races
+        // the main thread on Curtains' internal LazyThreadSafetyMode.NONE delegate. The
+        // listener must be registered before the decorViews.length == 0 early return,
+        // otherwise cold start (no root view yet) never registers a listener at all.
         uiThreadHandler.post(() -> {
+            Curtains.getOnRootViewsChangedListeners().add((view, added) -> {
+                if (added) {
+                    viewDrawInterceptor.Intercept(new View[]{view});
+                } else {
+                    viewDrawInterceptor.removeIntercept(new View[]{view});
+                }
+            });
+
             View[] decorViews = Curtains.getRootViews().toArray(new View[0]);//WindowManagerSpy.windowManagerMViewsArray();
 
             // Check if decorViews is not empty before accessing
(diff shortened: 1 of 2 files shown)


cdillard-NewRelic (approved):
:+1:

PR #631, opened 2026-09-16
fix: flush mobile logs on the fatal exception path and stamp per-entry sessionId (NR-616897)

ndesai-newrelic:
🔗 Linked to JIRA ticket: [NR-616897](https://newrelic.atlassian.net/browse/NR-616897)

## Jira

[NR-616897](https://new-relic.atlassian.net/browse/NR-616897)

## Headline

Flush Mobile Logs on the fatal-exception path, and stamp each log entry with its originating `sessionId`.

## What & Why

A session that ended in a fatal JS/Java error lost its Mobile Logs. The session's `MobileJSError` and its Session Replay both survived (different code paths), so the result in the UI was **a replay that exists but whose logs panel is empty** — reported externally as "session replay on JS errors is missing logs". In production (relibank-mobile-android, 30 days) 6 of 7 fatal-error sessions had **zero** `Log` records, while every non-fatal session had 20–70.

There are **two in-memory buffers in series** on the log path. The ticket identified only the second:

```
NewRelic.logInfo()
  → RemoteLogger executor QUEUE          ← loss window #1 (not in the ticket)
  → LogReporter.appendToWorkingLogfile()
  → BufferedWriter (8 KB)                ← loss window #2 (the ticket's root cause)
  → disk
```

Log lines only reach disk at harvest (`processLogs()`), `finalizeWorkingLogfile()`, or a payload roll. A fatal exception terminates the process before any of those run, so unflushed lines die with it — and because nothing reached disk, the leftover-file recovery in `onHarvestConnected()` has nothing to pick up on the next launch either. At a measured ~134 bytes/line, the 8 KB buffer flushes only once per ~61 lines, which is the size of that window.

`LogReporter` was the **only** append-without-flush path in the codebase. Every other store is write-through — `SessionReplayFileManager` uses the identical `AtomicReference<BufferedWriter>` + executor shape and flushes on every append, and `AbstractFileStore.writeAtomic()` does write+flush+close+rename per item.

**This PR fixes it on the fatal-exception path:** the crash handler now drains the logger queue (with a hard deadline) and then flushes the working logfile, synchronously, before it chains to the previous handler.

**Second, independent defect fixed here too — session attribution.** `getCommonBlockAttributes()` supplies one shared `sessionId` for the whole payload, resolved at *rollup* time. A file recovered on the next launch was therefore stamped with the **new** session's id, reproducing the same "replay with no logs" symptom for a different reason. Each entry now carries its own `sessionId`, which overrides the common block. Without this, fixing the flush alone would not have closed the customer report.

## Callouts

**This modifies the crash handler.** `UncaughtExceptionHandler.uncaughtException()`'s chain to the previous handler is load-bearing for any other crash reporter in the host app. The new block is therefore:
- the **first statement of the `finally`**, so it runs even if `storeAndReportCrash()` throws;
- **before** `PayloadController.shutdown()` and `chainExceptionHandler()`, since chaining usually never returns;
- wrapped in `catch (Throwable)` — not `IOException` — so a flush failure can never break the chain. `UncaughtExceptionHandlerTest` (9 pre-existing tests) still passes, and a new test asserts the chain explicitly.

**Order is load-bearing.** Drain the queue *first*, then flush the file. Reversed, queued entries never reach the buffer and the fix silently does nothing.

**`flush()` was unbounded.** The pre-existing `RemoteLogger.flush()` had no deadline — `executor.wait()` releases the monitor, so concurrent submissions could keep the queue non-empty indefinitely. On the crash path that would have blocked the handler chain. Added a bounded `flush(long)` overload; the no-arg method is untouched and keeps its exact prior semantics for its **13** existing test callers.

**Data queries:** `Log` records now carry a per-entry `sessionId` attribute. This is additive and overrides the common block. It adds ~50 bytes/line, accounted for by the existing `payloadBudget` arithmetic against the 1 MB Vortex limit. Restoring the commented-out `putAll(getCommonBlockAttributes())` was deliberately **not** done — it would have duplicated `entity.guid` plus every session attribute onto every line.

**Only `flush()`, never `fsync()`.** Once `write(2)` returns, bytes live in the kernel page cache, which survives process death by any means. `fsync` measured ~10× the cost for durability we don't need (kernel panic / power loss).

**Not covered — follow-up ticket needed.** ANR, NDK/native crash, OOM/low-memory-killer and force-stop all end in `SIGKILL`, which is uncatchable, so no handler runs. Covering those needs either a flush per append (measured +2.16 µs/line) or an mmap'd append log. `onNativeCrash` cannot help — it runs on the *next launch*, when the buffer is already gone.

**Also worth a follow-up:** recovered files still take non-`sessionId` attributes from the relaunched session. The complete fix is the pattern native crashes already use — `NativeReporting` resolves the origin session's frozen attributes from `SessionContextStore` via `lookupSessionManifest()`.

**Incidental finding (not changed here):** in `getCommonBlockAttributes()`, the live `getSessionID()` read is immediately clobbered by `attrs.putAll(sessionAttributes)`, because `AnalyticsAttribute.SESSION_ID_ATTRIBUTE` and `LogReporting.LOG_SESSION_ID` are both the literal `"sessionId"`. That live read is dead code. `Harvest.startSession()` is the only production path that rotates the session id, and it updates both `AgentConfiguration` and the cached analytics attribute together.

### New unit tests to review

| Test | Asserts |
|---|---|
| `LogReporterFlushTest.testFlushLandsLineOnDiskWithoutHarvestOrFinalize` | line reaches disk with no harvest/finalize. **Asserts the negative first** (not on disk *before* the flush) so it can't pass vacuously |
| `LogReporterFlushTest.testFlushIsSafeWhenWriterIsAlreadyClosed` | flushing a closed writer does not throw |
| `LogReporterFlushTest.testRecoveredLogKeepsOriginatingSessionIdAfterRotation` | after a rotation, the common block carries the **new** id while the record keeps its **originating** id — demonstrates both halves of the mechanism |
| `CrashReporterTest.testCrashPathFlushesLogsAndStillChainsToPreviousHandler` | queued + buffered lines reach disk on the crash path **and** the previous handler is still chained |
| `RemoteLoggerTest.testBoundedFlushReturnsWithinBudgetWhenQueueIsBlocked` | bounded drain honours its deadline with the pool saturated |
| `RemoteLoggerTest.testLogEntryCarriesItsOwnSessionId` | each entry carries its own `sessionId` |

### Manual verification

A **release** build is required — in a debug build a fatal JS error shows the RedBox and the process survives, so the next harvest flushes normally and the logs appear regardless.

1. Instrument a React Native app with `loggingEnabled: true`, Mobile Logs at 100% server-side.
2. Button A emits 16 identifiable lines: `NewRelic.logInfo('REPRO line N')`.
3. Button B performs an uncaught `throw new Error('REPRO fatal')`.
4. Build and install a **release** APK.
5. **Control:** launch, tap A, background via HOME, wait one harvest cycle.
6. **Crash:** force-stop, relaunch for a fresh session, tap A, tap B, confirm the process died, relaunch so the persisted error uploads.
7. Query both sessions:

```sql
SELECT sessionId, errorMessage, isFatalError, hasReplay FROM MobileJSError
WHERE appId = <APP_ID> WHERE isFatalError IS true SINCE 1 hour ago

SELECT count(*) FROM Log
WHERE entity.guid = '<GUID>' WHERE sessionId = '<SESSION_ID>' SINCE 1 hour ago
```

(`entity.guid` is stored unpadded — no trailing `==`.)

**Expected:** the fatal session returns its 16 lines, matching the control. Before this change it returned 0 of 16 while carrying `hasReplay: true` and a valid `MobileJSError`.

### Also worth confirming in review

- The bounded drain's 250 ms budget is a ceiling to bound the crash path, not a tuned value.
- `RemoteLoggerTest`'s bounded-flush test keeps 8×2 s blocking tasks deliberately — shortening them below the 1500 ms assertion bound would make the test pass even with the fix reverted. `shutdownNow()` in a `finally` stops them leaking non-daemon threads.

🤖 Generated with [Claude Code](https://claude.com/claude-code)

[NR-616897]: https://new-relic.atlassian.net/browse/NR-616897?atlOrigin=eyJpIjoiNWRkNTljNzYxNjVmNDY3MDlhMDU5Y2ZhYzA5YTRkZjUiLCJwIjoiZ2l0aHViLWNvbS1KU1cifQ
[NR-616897]: https://new-relic.atlassian.net/browse/NR-616897?atlOrigin=eyJpIjoiNWRkNTljNzYxNjVmNDY3MDlhMDU5Y2ZhYzA5YTRkZjUiLCJwIjoiZ2l0aHViLWNvbS1KU1cifQ


diegomtz5 (approved):
It does fix the bug, looks good!

PR #636, opened 2026-09-23
chore: add back create_release.yml

cmcadams-newrelic:
Adds back create_release.yml


ywang-nr (approved):

ywang-nr:
Take a look why the sonatype release workflow is failing

